Issue #207, opened 2024-12-30
Create an index for md files


tkpratardan:
Goal: Generate an index file listing all markdown files in the repositor

Steps:
- List all markdown files and organize them (hierarchically )
- Include metadata:
     - File name.
     - Relative path.

also handle #210

gpsaggese:
Yes just re-iterating to avoid confusion / frustration.
1. Apply the changes in the PR requested by the previous reviewer. In this way you can get familiar with the code
2. Then we can decide where to go from here
    - We could summarize each doc with ChatGPT (we have a pipeline for doing that)

On the other side, the project of ChatGPT-ify the documentation might make this issue superfluous. In any case let's see where we are and then we'll decide.

@aangelo9 the way of doing things is to always think hard, make a proposal, and after the plan is clear and agreed upon, execute

aangelo9:
Thanks for the clarification.

I'll go ahead and make the changes, and once that's done, I will reasses and make a proposal.

aangelo9:
I have resolved most of the requested changes. The remaining requests needs the code to run.

I have added a unit test that is not ran yet because it calls openai. I just want to confirm if I am allowed to run it since it could cost money, or should I just run a mock test, so it omits all openai calls. 

I also want to propose a change in the code:

- Instead of taking the git root repo and processing everything in it, we could add an argument to a specific folder and process things per folder instead of an entire repo. This makes testing easier as well.

sonniki:
> I have resolved most of the requested changes. The remaining requests needs the code to run.

When you think it's ready, go ahead and convert the PR from draft to "ready for review" and re-request review through the GH interface.

> I have added a unit test that is not ran yet because it calls openai. I just want to confirm if I am allowed to run it since it could cost money, or should I just run a mock test, so it omits all openai calls.

There is a conversation about it in the PR. See my proposal in https://github.com/causify-ai/helpers/pull/237#issuecomment-2607306977.

 > Instead of taking the git root repo and processing everything in it, we could add an argument to a specific folder and process things per folder instead of an entire repo. This makes testing easier as well.

Sounds good. Maybe the repo root can be the default value for the argument.

aangelo9:
I decided to remove `--generate_summary` and `--update_summary` args. So the code now just refreshes the README index for all markdown files.
Also added `--use_placeholder_summary` argv, to bypass openai usage and conduct mock tests.

[...]

PR #237, opened 2025-01-15
HelpersTask207_create_index_for_readme

tkpratardan:
#207


linters/generate_readme_index.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    Generate a two-line summary for the given Markdown file.
+
+    :param file_path: Path to the Markdown file
+    :return: A generated summary string
+    """
+    content = hio.from_file(file_path)
+    prompt = f"Summarize the following content into two lines:\n\n{content}"
+    summary = hopenai.get_completion(user=prompt, model="gpt-4o-mini")
+    return summary.strip()
```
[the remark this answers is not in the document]
aangelo9: If openai is used, prompt is fortified to guarantee generation of only 2 sentences.
```
prompt = (
            "Summarize the following content in exactly two lines. "
            "Do not include any introduction or list markers. "
            "Just return the summary itself, nothing else.\n\n"
            f"{content}"
        )
```

aangelo9:
`import openai` raised an error 'No Module Found', which means that the docker container do not have openai installed. I also do not have the authority to install inside docker. So a temp fix is to encapsulate `import hopenai` in the function where it is called `generate_summary_for_file()`.

Logic changes are in https://github.com/causify-ai/helpers/issues/207#issuecomment-2795737604.

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,199 @@
...
+
+    _LOG.debug("Generating real summary for: %s", file_path)
+    content = hio.from_file(file_path)
+    prompt = (
+        "Summarize the following content in exactly two lines. "
+        "Do not include any introduction or list markers. "
+        "Just return the summary itself, nothing else.\n\n"
+        f"{content}"
+    )
+    summary = hopenai.get_completion(user_prompt=prompt, model="gpt-4o-mini")
```
tkpratardan, line 101: I would probably want control on the choice of model probably as option because the default model here is not good enough for longer context lengths. 
@sonniki can take a call.

tkpratardan (commented):
Some style checks. 
FYR: https://github.com/causify-ai/helpers/blob/master/docs/coding/all.coding_style.how_to_guide.md

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,199 @@
...
+    :return: a short summary string
+    """
+    if use_placeholder:
+        _LOG.debug("Using placeholder summary for %s", file_path)
+        file_name = os.path.basename(file_path)
+        summary = f"Placeholder summary for {file_name}"
+        return summary
+    # Prevents hopenai to be imported when not needed
+    import helpers.hopenai as hopenai
+
```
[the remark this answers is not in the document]
aangelo9: I mentioned this in https://github.com/causify-ai/helpers/pull/237#issuecomment-2795739338.
The docker env for i run_fast_tests do not have the import openai module installed, so tests will fail. Is there a different approach to this problem?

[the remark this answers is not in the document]
tkpratardan: I understand. I meant imports all together at the beginning of the script. This is not as per the style we follow.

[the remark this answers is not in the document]
aangelo9: Understood. I will follow style conventions. I do not have the ubuntu password to install inside the docker container. I have installed openai in my local client and the code works great.

I do have some suggestions to work around this.

- We could implement this into hopenai.py and/or generate_readme_index.py to guarantee the import. (Import will still fail in container but will be a failsafe on a different env).
```
# Taken from helpers/hchatgpt.py
try:
    import openai
except ImportError:
    os.system("pip install openai")
finally:
    import openai
```
- I see that there are also no tests for `helpers/hchatgpt.py` and `helpers/hopenai.py`. So unit tests should also be removed or converted to a test file like in `dev_scripts_helpers/chatgpt/run_simple_chatgpt.py`, so the tests are not included in run_fast_tests or run_slow_tests,  since the openai API should not be called on every fast_tests. 

I vote to just remove the test file since testing locally has no issue.

dev_scripts_helpers/documentation/generate_readme_index.py
sonniki: Agreed, we can expose the model as a parameter, and have `None` as the default value (since this is what is used in `hopenai.get_completion()` as default)

dev_scripts_helpers/documentation/generate_readme_index.py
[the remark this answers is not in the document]
sonniki: > We could implement this into hopenai.py and/or generate_readme_index.py to guarantee the import. (Import will still fail in container but will be a failsafe on a different env).

Let's 1) replicate the try-except-finally approach to importing `openai` in `hopenai.py`, as it is done in `hchatgpt.py`; 2) file an issue to run `openai` as a dockerized executable. @aangelo9 please file an issue and put me and @gpsaggese as FYI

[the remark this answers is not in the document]
sonniki: > I see that there are also no tests for helpers/hchatgpt.py and helpers/hopenai.py. So unit tests should also be removed or converted to a test file like in dev_scripts_helpers/chatgpt/run_simple_chatgpt.py, so the tests are not included in run_fast_tests or run_slow_tests, since the openai API should not be called on every fast_tests.
I vote to just remove the test file since testing locally has no issue.

I would still like to keep the tests, with the `use_placeholder` flag in order not to spam the API. If importing `openai` down the line causes problems for running tests, we can the `pytest.importorskip` decorator to skip tests if `openai` is not installed.

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,199 @@
...
+import helpers.hio as hio
+
+_LOG = logging.getLogger(__name__)
+
+
+def list_markdown_files(repo_path: str) -> Set[str]:
+    """
+    List all Markdown files in the given repository path.
+
+    :param `repo_path`: root path of the repository
```
sonniki, line 62: Why the backticks everywhere? Pls see how param/return descriptions are formatted in the codebase

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,199 @@
...
+    :return: a set containing the relative paths of all `.md` files found
```
sonniki, line 63: The convention is not to mention types in param/return descriptions - the fact that it's a set is clear from the typing hints

dev_scripts_helpers/documentation/generate_readme_index.py
sonniki, line 62: Haven't we switched from repo root to just any dir? So it should be now just `dir_path` everywhere, with optionally the repo root path as default

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+    Generate a two-line summary for a given Markdown file.
+
+    :param `file_path`: full path to the Markdown file
+    :param `use_placeholder`: use a dummy summary instead of calling OpenAI
+    :return: a short summary string
+    """
+    if use_placeholder:
+        _LOG.debug("Using placeholder summary for %s", file_path)
+        file_name = os.path.basename(file_path)
```
sonniki, line 87: I would keep the full path here

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,199 @@
...
+                f"- **File Name**: {filename}  \n"
+                f"  **Relative Path**: [{rel_path}]({rel_path})  \n"
+                f"  **Summary**: {summary}  \n"
+            )
+    updated_content = "\n".join(lines)
+    hio.to_file(readme_file_path, updated_content)
+    _LOG.debug("README updated at: %s", readme_file_path)
+
+
+def _main() -> None:
```
sonniki, line 170: I don't really understand why having two modes (generate vs. update) was problematic. Even though the idea is that this script runs once in a while, it's still going to take a looong time to run for all md files in the repo if it's run with openai summary generation, and if there is a way to run the script to only update the index with the new additions to the docs, it's useful.
To make it more intuitively clear, we can have an argument `mode` with the following options:
- `overwrite`: basically what you have here now, which is completely rewriting the README file from scratch
- `append`: keep what's already in the README, add only md files that aren't there yet

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,199 @@
...
+        action="store_true",
+        help="Use a placeholder summary instead of calling OpenAI.",
+    )
+    args = parser.parse_args()
+    repo_path = args.repo_path
+    use_placeholder = args.use_placeholder_summary
+
+    # Fetch all Markdown files in the repository.
+    markdown_files = list_markdown_files(repo_path)
+    # Check for Markdown file
```
sonniki, line 191: -> `Update Markdown index.`
and it should be moved below `if markdown_files:`

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,199 @@
...
+        else:
+            dir_display = os.path.basename(os.path.abspath(repo_path))
+        # Directory Header
+        lines.append(f"### {dir_display}")
+        lines.append("")
+        # Summary block generation
+        for filename, rel_path in sorted(directory_map[directory]):
+            summary = summaries.get(
+                rel_path, f"No summary available for {rel_path}."
+            )
```
sonniki, line 156: How is it possible not to have a summary?

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,199 @@
...
+        file_path = os.path.join(repo_path, rel_path)
+        summaries[rel_path] = generate_summary_for_file(
+            file_path, use_placeholder
+        )
+    # Build directory map
+    directory_map = defaultdict(list)
+    for md_file in markdown_files:
+        directory = os.path.dirname(md_file)
+        filename = os.path.basename(md_file)
+        directory_map[directory].append((filename, md_file))
```
sonniki, line 132: Maybe it's best to do it in `list_markdown_files` already

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,199 @@
...
+        "This section lists all Markdown files in the repository.",
+        "",
+    ]
+    # Create summary block
+    for directory in sorted(directory_map.keys()):
+        # Gets directory or root name
+        if directory:
+            dir_display = directory
+        else:
+            dir_display = os.path.basename(os.path.abspath(repo_path))
```
sonniki, line 148: This part can also go to the construction of the directory map

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,199 @@
...
+    lines = [
+        "# Repository README",
+        "",
+        "## Markdown Index",
+        "",
+        "This section lists all Markdown files in the repository.",
+        "",
+    ]
+    # Create summary block
+    for directory in sorted(directory_map.keys()):
```
sonniki, line 143: And sorting can be done there already. We just want to keep functions as distinct, compact and "clean" as possible. E.g. this one should be only for index generation

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,199 @@
...
+        if directory:
+            dir_display = directory
+        else:
+            dir_display = os.path.basename(os.path.abspath(repo_path))
+        # Directory Header
+        lines.append(f"### {dir_display}")
+        lines.append("")
+        # Summary block generation
+        for filename, rel_path in sorted(directory_map[directory]):
+            summary = summaries.get(
```
sonniki, line 154: We can also generate the summary right here -- constructing the file-to-summary mapping doesn't do much if we only use it once to look up the summary here.

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,111 @@
...
+import helpers.hio as hio
+import helpers.hunit_test as hunitest
+
+
+# #############################################################################
+# Test_generate_readme_index
+# #############################################################################
+
+
+class Test_generate_readme_index(hunitest.TestCase):
```
sonniki, line 14: Our approach is testing functions rather than whole scripts. So here we'd want to have two test classes: one for testing `generate_markdown_index()` and one for testing `list_markdown_files()`

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,111 @@
...
+        dshdgrein._main()
+        # Read
+        readme_path = os.path.join(repo_path, "README.md")
+        readme_content = hio.from_file(readme_path)
+        # Check
+        self.check_string(readme_content, tag="README.md")
+
+    def test3(self) -> None:
+        """
+        Test for REAME file generation on an empty directory.
```
sonniki, line 97: typo, README

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,111 @@
...
+        This markdown file is for testing the generate_readme_index.py script.
+
+        ## Introduction
+
+        It should be detected in the index and updated with a summary if needed.
+        """
+        file_name = "sample.md"
+        repo_path = self.get_scratch_space()
+        self.write_input_file(content, file_name)
+
```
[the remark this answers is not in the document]
sonniki: The PR to add a Linter step that removes empty lines is still under review. So for now we should just make sure to remove them manually.

dev_scripts_helpers/documentation/generate_readme_index.py
sonniki: Instead of `append`, we can do `refresh`: add new md files that weren't in the readme yet + delete references to the files that are not in the repo anymore (because they got deleted or renamed)

dev_scripts_helpers/documentation/generate_readme_index.py
aangelo9: I will revert back to using no backticks. The comment https://github.com/causify-ai/helpers/pull/237#discussion_r1925207396 is probably outdated.

sonniki: Oh, I see:) I understand the confusion now - this actually referred to "repo_path" in the param description: "markdown file paths (relative to repo_path)"

dev_scripts_helpers/documentation/generate_readme_index.py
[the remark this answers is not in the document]
aangelo9: There are some proposed solutions on https://github.com/causify-ai/helpers/issues/539#issuecomment-2800579673. WDYT?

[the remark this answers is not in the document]
sonniki: IMO just using the `use_placeholder` flag + `pytest.importorskip` on tests is fine for now

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
+import os
+import textwrap
+
+import pytest
+
+# Equivalent to `import openai`, but skip this module if the module is not present.
+# `mockai` must be imported before `openai` to properly mock it.
+mockai = pytest.importorskip("openai")
+
+# Pylint wrong-import-position
```
sonniki, line 10: Is there a way to silence `pylint`'s complaint about the wrong import position? E.g. there are sometimes "code phrases" in such libs to do so

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
...
+# It is necessary that generate_readme_index is imported after mockai.
+# If not, real OpenAI API will be called.
```
sonniki, line 12: Then we would be able to remove this.

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+# #############################################################################
+# Test_list_markdown_files
+# #############################################################################
+
+
+class Test_list_markdown_files(hunitest.TestCase):
+
+    def write_input_file(self, txt: str, file_name: str) -> str:
```
sonniki, line 25: Better to make it a private method (`_write_input_file`)
Same for similar methods in other test classes

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+        Write test content to a file in the scratch space.
+
+        :param testcase: instance of the test case (self)
```
sonniki, line 29: We don't add this to the param descriptions

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+        Test for list generation of directory.
+        """
+        # Sample nested documents.
+        file_structure = {
+            "welcome.md": "# welcome page",
+            "docs/intro.md": "# Introduction",
+            "docs/guide/setup.md": "# Setup Guide",
+            "docs/guide/usage.md": "# Usage Guide",
+        }
```
sonniki, line 53: I would also add here some non-Markdown files

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
...
+        dir_name = self.get_scratch_space()
+        file_path = os.path.join(dir_name, file_name)
+        file_path = os.path.abspath(file_path)
+        # Create the file.
+        hio.to_file(file_path, txt)
+        return file_path
+
+    def test1(self) -> None:
+        """
+        Test for list generation of directory.
```
sonniki, line 45: Test retrieving all Markdown files in a directory.

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
...
+        for path, content in file_structure.items():
+            self.write_input_file(content, path)
+        # Run.
+        output = dshdgrein.list_markdown_files(dir_path)
+        # Check.
+        self.assertEqual(output, expected)
+
+    def test2(self) -> None:
+        """
+        Test for ignore existing README.
```
sonniki, line 71: Test that the existing README is ignored

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
...
+            "docs/guide/setup.md",
+            "docs/guide/usage.md",
+            "docs/intro.md",
+            "welcome.md",
+        ]
+        dir_path = self.get_scratch_space()
+        for path, content in file_structure.items():
+            self.write_input_file(content, path)
+        # Run.
+        output = dshdgrein.list_markdown_files(dir_path)
```
sonniki, line 65: By convention, we call it `actual`. Pls change everywhere

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
...
+# #############################################################################
+# Test_generate_readme_index
+# #############################################################################
+
+
+class Test_generate_readme_index(hunitest.TestCase):
+
+    def write_readme(self, content: str) -> str:
+        """
+        Creates a README file with content.
```
sonniki, line 112: -> "Create"

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
...
+        ]
+        # Run.
+        output = dshdgrein.generate_markdown_index(
+            readme_path=readme_path,
+            markdown_files=markdown_files,
+            index_mode="generate",
+            model="placeholder",
+        )
+        # Check.
+        self.check_string(output, tag="README.md")
```
sonniki, line 144: We generally don't use this `tag` param

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+        content = textwrap.dedent(content)
+        dir_path = self.get_scratch_space()
+        readme_path = os.path.join(dir_path, "README.md")
+        hio.to_file(readme_path, content)
+        return readme_path
+
+    def test1(self) -> None:
+        """
+        Test for generating README from scratch using placeholder summary.
```
sonniki, line 125: Remove "for"

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
...
+            "docs/guide/usage.md",
+            "docs/intro.md",
+            "welcome.md",
+        ]
+        # Run.
+        output = dshdgrein.generate_markdown_index(
+            readme_path=readme_path,
+            markdown_files=markdown_files,
+            index_mode="generate",
+            model="placeholder",
```
sonniki, line 141: By convention, let's move 
```
index_mode="generate"
model="placeholder"
```
after declaring `markdown_files` (so that all the params are declared at one place), and then here do
```
index_mode=index_mode,
model=model,
```

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
...
+        )
+        # Check.
+        self.check_string(output, tag="README.md")
+
+    def test2(self) -> None:
+        """
+        Test for refresh README to add new file using placeholder summary.
```
sonniki, line 148: "Test refreshing README by adding a new file."

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
...
+            markdown_files=markdown_files,
+            index_mode="refresh",
+            model="placeholder",
+        )
+        # Check.
+        self.check_string(output, tag="README.md")
+
+    def test3(self) -> None:
+        """
+        Test for refresh README to remove summary of deleted file.
```
sonniki, line 196: "Test refreshing README by removing an obsolete file."

dev_scripts_helpers/documentation/test/test_generate_readme_index.py
```diff
@@ -0,0 +1,281 @@
...
+            markdown_files=markdown_files,
+            index_mode="refresh",
+            model="placeholder",
+        )
+        # Check.
+        self.check_string(output, tag="README.md")
+
+    def test4(self) -> None:
+        """
+        Test for refresh README to add and delete file.
```
sonniki, line 238: "Test refreshing README by adding a new file and removing another."

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,250 @@
...
+        help="Path to target folder. Defaults to Git root.",
+    )
+    parser.add_argument(
+        "--model",
+        type=str,
+        required=True,
+        help="Choice of ChatGPT model. Use `placeholder` to create dummy summary.",
+    )
+    args = parser.parse_args()
+    # Fetch all Markdown files in the repository.
```
sonniki, line 232: Also here -- not repository, since it can be just one dir now.

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,250 @@
...
+    hierarchically by directory, and appends this index to the README
+    along with basic metadata.
+
+    :param readme_path: path to README.md file to read/write
+    :param markdown_files: list of relative Markdown paths
+    :param index_mode: whether to "generate" or "refresh" summaries
+    :param model: summarization model to use
+    :return: complete Markdown index content
+    """
+    summaries = {}
```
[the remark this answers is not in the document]
aangelo9: I found that `dev_scripts_helpers/aws/am_aws.py` has a similar situation.
```
parser.add_argument(
        "--index_mode",
        choices=["generate", "refresh"],
        required=True,
        help="Choose execute index mode: generate or refresh",
    )
```
Since the parser already guards the inputs and raises the error if input is not in the listed choices, the final `else: raise error` is unnecessary. I will implement if-elif for better clarity if more options are implemented in the future.

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,250 @@
...
+    parser.add_argument(
+        "--dir_path",
+        type=str,
+        default=hgit.find_git_root(),
+        help="Path to target folder. Defaults to Git root.",
+    )
+    parser.add_argument(
+        "--model",
+        type=str,
+        required=True,
```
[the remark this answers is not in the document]
aangelo9: That is true. I will default the value to use the placeholder to prevent accidentally using real LLM models.

dev_scripts_helpers/documentation/generate_readme_index.py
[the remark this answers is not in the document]
sonniki: > Since the parser already guards the inputs and raises the error if input is not in the listed choices, the final else: raise error is unnecessary.

This is true, but the function `generate_markdown_index()` can in theory be used on its own too, not in the script - e.g., in tests, but possibly also outside of test environment. 

> I will implement if-elif for better clarity if more options are implemented in the future.

Yes, in any case it brings more clarity by making these possible values more "visible".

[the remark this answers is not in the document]
aangelo9: Understood, I will implement `else: raise error` for when `generate_markdown_index()` is used outside of the main script.

dev_scripts_helpers/documentation/generate_readme_index.py
```diff
@@ -0,0 +1,250 @@
...
+    file_path: str, *, model: Optional[str] = None
+) -> str:
+    """
+    Generate a two-line summary for a given Markdown file.
+
+    :param file_path: full path to the Markdown file
+    :param model: model identifier (e.g., "placeholder", "gpt-4o-mini")
+    :return: a short summary of a file
+    """
+    if model is None:
```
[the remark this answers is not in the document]
aangelo9: On https://github.com/causify-ai/helpers/pull/237#discussion_r2049470596. Since the function `generate_markdown_index()` could be used outside of the script, it is possible for model to be None. I could either use the placeholder summary if `model=None`, or let it pass through `hopenai.get_completion` since it default models to "gpt-4o-mini" when `model=None`.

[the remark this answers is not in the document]
sonniki: Fair enough. But we could make it default to `placeholder` for example, not to None. Similarly to how the default value of the script arg is `placeholder`.

sonniki (approved):

Issue #307, opened 2025-02-21
Fix docformatter


sonniki:
Still exhibiting buggy behavior on this docstring (from `helpers/lib_tasks_lint.py`):

```
def lint(  # type: ignore
    ctx,
    stage="prod",
    version="",
    files="",
    dir_name="",
    modified=False,
    last_commit=False,
    branch=False,
    only_format=False,
    only_check=False,
    no_dev_server=False,
):
    """
    Lint files.

    ```
    # To lint specific files:
    > i lint --files="dir1/file1.py dir2/file2.md"

    # To lint all the files in the current dir using only formatting actions:
    > i lint --dir-name . --only-format

    # To lint the files modified in the current git client:
    > i lint --modified

    # To exclude certain paths from linting:
    > i lint --files="$(find . -name '*.py' -not -path './compute/*' -not -path './amp/*')"
    ```

    :param stage: the image stage to use (e.g., "prod", "dev", "local")
    :param version: the version of the container to use
    :param files: specific files to lint (e.g. "dir1/file1.py dir2/file2.md")
    :param dir_name: name of the dir where all files should be linted
    :param modified: lint the files modified in the current git client
    :param last_commit: lint the files modified in the previous commit
    :param branch: lint the files modified in the current branch w.r.t. master
    :param only_format: run only the modifying actions of Linter (e.g., black)
    :param only_check: run only the non-modifying actions of Linter (e.g., pylint)
    :param no_dev_server: True, if running the linter on local machine,
        else false if on dev server
    """
```

After linting the "param" section is pushed to the right.

sonniki:
Done

PR #308, opened 2025-02-21
HelpersTask307: Fix docformatter

sonniki:
#307 
Fixing the indent bug + adding tests

### DIFF
--- a/linters/test/test_amp_doc_formatter.py
+++ b/linters/test/test_amp_doc_formatter.py
@@ -190,6 +190,21 @@ def long_docstring_line(cmd: str) -> None:
     > git pull
     ```
     """
+
+def empty_lines_in_code_block(cmd: str) -> None:
+    """
+    Test docstring.
+
+    ```
+    # To lint the files modified in the current git client:
+    > i lint --modified
+
+    # To exclude certain paths from linting:
+    > i lint --files="$(find . -name '*.py' -not -path './compute/*' -not -path './amp/*')"
+    ```
+
+    Text after.
+    """
 '''
         expected = text
         actual = self._docformatter(text)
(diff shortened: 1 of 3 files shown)


linters/test/test_amp_doc_formatter.py
```diff
@@ -190,6 +190,21 @@ def long_docstring_line(cmd: str) -> None:
...
+
+    # To exclude certain paths from linting:
+    > i lint --files="$(find . -name '*.py' -not -path './compute/*' -not -path './amp/*')"
+    ```
+
+    Text after.
+    """
 '''
         expected = text
         actual = self._docformatter(text)
```
gpsaggese, line 210: can we use self.get_scratch instead of temporary files?
You can file an outsourceable bug and we can get an intern to do it

linters/test/test_amp_doc_formatter.py
```diff
@@ -190,6 +190,21 @@ def long_docstring_line(cmd: str) -> None:
     > git pull
     ```
     """
+
```
gpsaggese, line 193: Should we put this in a file?
It's becoming a huge Python file inside a Python file...

gpsaggese (commented):
Good work. 2 small follow ups (you can file bugs)

linters/test/test_amp_doc_formatter.py
sonniki: https://github.com/causify-ai/helpers/issues/330

linters/test/test_amp_doc_formatter.py
sonniki: https://github.com/causify-ai/helpers/issues/331

Issue #296, opened 2025-02-18
Run Linter only on certain file types


sonniki:
Getting this error in the [GH actions run](https://github.com/causify-ai/helpers/actions/runs/13400933093/job/37431387584?pr=295):
```
RuntimeError: error='utf-8' codec can't decode byte 0x89 in position 0: invalid start byte
file_name='docs/work_organization/figs/use_github/Draft_PR.png'
```

Apparently, nothing is stopping a PNG file from getting to Linter scripts, which then fail because they cannot decode it. Investigate at what point this error occurs exactly (I'd thought all of the individual Linter scripts had a condition to skip a file if it is not Python/Markdown/etc) and fix.

sonniki:
All done

gpsaggese:
Added a couple of comments in https://github.com/causify-ai/helpers/pull/325/files

Should we skip based on type file and then use the RuntimeError with warning if it's a valid file?
It seems that this is a general problem that warrants a general solution.

Re-opening to track

sonniki:
> Should we skip based on type file and then use the RuntimeError with warning if it's a valid file?
It seems that this is a general problem that warrants a general solution.

In those two Linter scripts we want to apply the action on all file types that can be "interpreted" as text and therefore opened with `hio.from_file()`. So it's not possible to skip all file types except a select few. On the other hand, I doubt we can explicitly list all the possible file types that do not satisfy the conditions. I agree that using `try-except` is hacky but IMO it's more robust in this case. WDYT?

> Added a couple of comments in https://github.com/causify-ai/helpers/pull/325/files

Answered there.

gpsaggese:
Sounds good. Thanks

PR #325, opened 2025-02-27
HelpersTask296: Add workaround for files that cannot be read

sonniki:
#296 Fix the bug that appeared when Linter encountered non-txt-based files (e.g. png)

### DIFF
--- a/linters/amp_check_merge_conflict.py
+++ b/linters/amp_check_merge_conflict.py
@@ -34,6 +34,11 @@ def _check_merge_conflict(file_name: str, line_num: int, line: str) -> str:
     return msg
 
 
+# #############################################################################
+# _CheckMergeConflict
+# #############################################################################
+
+
 class _CheckMergeConflict(liaction.Action):
     """
     Check if there is a git merge conflict marker in the file.
@@ -44,8 +49,12 @@ class _CheckMergeConflict(liaction.Action):
 
     def _execute(self, file_name: str, pedantic: int) -> List[str]:
         _ = pedantic
-        # Read the input file.
-        lines = hio.from_file(file_name).split("\n")
+        try:
+            # Read the input file.
+            lines = hio.from_file(file_name).split("\n")
+        except RuntimeError:
+            _LOG.debug("Cannot read file_name='%s'; skipping", file_name)
+            return []
         output = []
         # Check the file lines for merge conflict markers.
         for i, line in enumerate(lines):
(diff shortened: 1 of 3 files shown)


linters/amp_check_merge_conflict.py
```diff
@@ -44,8 +49,12 @@ def check_if_possible(self) -> bool:
 
     def _execute(self, file_name: str, pedantic: int) -> List[str]:
         _ = pedantic
-        # Read the input file.
-        lines = hio.from_file(file_name).split("\n")
+        try:
+            # Read the input file.
+            lines = hio.from_file(file_name).split("\n")
+        except RuntimeError:
+            _LOG.debug("Cannot read file_name='%s'; skipping", file_name)
```
gpsaggese, line 56: Should this be a _LOG.warning?

linters/amp_fix_whitespaces.py
```diff
@@ -79,8 +84,12 @@ def check_if_possible(self) -> bool:
 
     def _execute(self, file_name: str, pedantic: int) -> List[str]:
         _ = pedantic
-        # Read the input file.
-        lines = hio.from_file(file_name).split("\n")
+        try:
```
gpsaggese, line 87: Same Q about _LOG.warning

Also if it's an idiom we can factor it out in a method of the class itself, so everybody can uses it.

sonniki: It is indeed an idiom. Every Linter scripts of ours has something like this:
```
        if not liutils.is_py_file(file_name) and not file_name.endswith(".ipynb"):
            # Applicable only to Python files and Ipynb notebooks.
            _LOG.debug("Skipping file_name='%s'", file_name)
            return []
```
I'll file a bug to factor it out as a method (can be a good "onboarding bug" for interns).

linters/amp_check_merge_conflict.py
sonniki: Well, I'm not sure. In this case it's not a "problem" that a file cannot be read, it is expected behavior for some file types. So IMO there is no need for a warning -- there is nothing that should be fixed -- it's just debug info that certain Linter actions do not apply to certain file types.

linters/amp_fix_whitespaces.py
sonniki: https://github.com/causify-ai/helpers/issues/332

Issue #348, opened 2025-03-10
helpers:prod error on GH actions


sonniki:
FYI @samarth9008 @gpsaggese 

Recent GH actions that use the `helpers:prod` image are failing:

```
 hsystem.py _system:261             
################################################################################
cmd='(docker inspect ***.dkr.ecr.eu-north-1.amazonaws.com/helpers:prod --format '{{.Architecture}}') 2>&1' failed with rc='1'
################################################################################
Output of the failing command is:
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

Error: No such object: ***.dkr.ecr.eu-north-1.amazonaws.com/helpers:prod
```

Examples of failed jobs:
https://github.com/causify-ai/helpers/actions/runs/13764472696/job/38487668244
https://github.com/causify-ai/cmamp/actions/runs/13753850192/job/38458225773

samarth9008:
Feel free to provide your solution as welll

dremdem:
> From what I can see is, there is no code for pulling the image.

I found the same, but wondering how it was working before? 

For example 2 days ago: 
https://github.com/causify-ai/helpers/actions/runs/13733817172/job/38415030296

dremdem:
I'll try to ssh into GH container and figure out.

samarth9008:
> Still confused how it was working before.

Same question. I thought the system would be pulling the image if not found locally but now that might not be the case.

> Will try to add a hack, 

Instead of hack, I think it can be a proper solution. That step is already done for other types of run. We just need to copy and paste. Make sure we do it across all the repos.

sonniki:
Note that not only the Linter workflow is failing but also the import cycle detector (which also uses `helpers:prod`)
- Failed run example: https://github.com/causify-ai/cmamp/actions/runs/13753850192/job/38458225773
- Workflow link: https://github.com/causify-ai/cmamp/blob/master/.github/workflows/import_cycles_detector.yml

samarth9008:
@dremdem 

Regarding this comment - https://github.com/causify-ai/helpers/issues/348#issuecomment-2710737644

There are few problems with import cycle detector run

- When ran from `cmamp` it pulls cmamp dev image which is wrong. It should pull `helpers:prod` image. Maybe we need to change the pull image steps and should use invoke `invoke docker_pull_helpers`. Check the invoke once.

[...]

PR #349, opened 2025-03-10
HelpersTask348_helpersprod_error_on_GH_actions

dremdem:
#348

- check lint error

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -58,7 +58,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r .github/gh_requirements.txt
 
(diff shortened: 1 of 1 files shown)


.github/workflows/linter.yml
```diff
@@ -79,6 +79,7 @@ jobs:
         # To run the linter, fetch the master branch for comparison with the
         # current branch.
         run: |
+          invoke docker_pull_helpers
```
dremdem, line 82: Not sure, it might be better to make it as a separate step. 
WDYT @samarth9008?

samarth9008: I would add a pull image step as we do it on the fast/slow.. test

PR #353, opened 2025-03-10
CmampTask11449_Debug_why_failing_tests_do_not_cause_the_GitHub_Actions_pipeline_to_fail

heanhsok:
Task https://github.com/causify-ai/cmamp/issues/11449


main_pytest.py
```diff
@@ -92,10 +93,14 @@ def _run_test(runnable_dir: str, command: str) -> None:
     # devops and helpers directory.
     env = os.environ.copy()
     env["HELPERS_ROOT_DIR"] = os.path.join(os.getcwd(), "helpers_root")
-    env[
-        "PYTHONPATH"
-    ] = f"{os.path.join(os.getcwd(), runnable_dir)}:{env['HELPERS_ROOT_DIR']}"
-    subprocess.run(f"invoke {command}", shell=True, env=env, cwd=runnable_dir)
+    env["PYTHONPATH"] = (
```
gpsaggese, line 96: Ok but let's explain why we don't use hsystem.system and what is the problem / solution (e.g., a reference to the bug is enough).

Without comments, the reader will wonder "why this weird code? Let me simplify it"

Issue #401, opened 2025-03-20
External linux check triggered by running docker


sonniki:
As reported in https://github.com/causify-ai/helpers/issues/386#issuecomment-2740974001 and https://github.com/causify-ai/helpers/issues/391#issuecomment-2741307266, when a user working on Mac executes commands that run a Docker container (`i lint`, `i docker_bash`, etc), they get an error because both `is_mac()` and `is_external_linux()` end up being `True`.

```
(client_venv.helpers) (base) Indrayudds-MacBook-Air:helpers1 indro$ i lint --files="dir1/file1.py dir2/file2.py"
18:13:44 - INFO  hdbg.py init_logger:1018                               > cmd='/Users/indro/src/venv/client_venv.helpers/bin/invoke lint --files=dir1/file1.py dir2/file2.py'
# lint: stage='prod', version='', files='dir1/file1.py dir2/file2.py', dir_name='', modified=False, last_commit=False, branch=False, num_threads='-1', only_format=False, only_check=False
18:13:44 - INFO  lib_tasks_lint.py lint:260                             All Linter actions selected
18:13:44 - WARN  hserver.py _raise_invalid_host:342                     Don't recognize host: host_os_name=Darwin, am_host_os_name=None
IMAGE=causify/helpers:prod \
        docker compose \
        --file /Users/indro/src/helpers1/devops/compose/docker-compose.yml \
        --env-file devops/env/default.env \
        run \
        --rm \
        --name indro.helpers.app.helpers1.20250320_141344 \
        --user $(id -u):$(id -g) \
        app \
        $(find . -path '*linters/base.py') --files dir1/file1.py dir2/file2.py --num_threads -1 
time="2025-03-20T14:13:44-04:00" level=warning msg="The \"CSFY_FORCE_TEST_FAIL\" variable is not set. Defaulting to a blank string."
time="2025-03-20T14:13:44-04:00" level=warning msg="The \"CSFY_AWS_ACCESS_KEY_ID\" variable is not set. Defaulting to a blank string."
time="2025-03-20T14:13:44-04:00" level=warning msg="The \"CSFY_AWS_DEFAULT_REGION\" variable is not set. Defaulting to a blank string."
time="2025-03-20T14:13:44-04:00" level=warning msg="The \"CSFY_AWS_SECRET_ACCESS_KEY\" variable is not set. Defaulting to a blank string."
[...]

PR #404, opened 2025-03-21
HelpersTask401_External_linux_check_triggered_by_running_docker

dremdem:
#401

- change `is_external_linux()` to consider:
   - inside the Docker container


sonniki (commented):
LGTM
@indrayudd @allenmatt10 could any of you switch to this branch and try the commands that were causing the error before? Let's confirm that the fix works.

allenmatt10:
I've tried out the commands, and can confirm the fix works on this branch. Thanks for the update.

Issue #380, opened 2025-03-15
Unit test extract_notebook_images


gpsaggese:
Add unit test for https://github.com/causify-ai/tutorials/issues/57

We want to have a notebook that we process using our usual approach of `hunitest.TestCase`.

Let's test end-to-end. We want to check that the output is correct

Shaunak01:
@gpsaggese testing the dockerized executable will build the container from scratch every time, wont it take a long time for the test to run then? How are we supposed to go about this? or just test the independent logic? (which i am not sure is applicable in this scenario)

gpsaggese:
Good question.

1) Sometimes we can find almost built containers to build on top, rather than building from scratch
2) If it's too slow to be a fast/slow tests, we can make those tests super-slow

Shaunak01:
PR: https://github.com/causify-ai/helpers/pull/422

PR #422, opened 2025-03-24
HelpersTask380_Unit_test_extract_notebook_images

Shaunak01:
Issue: https://github.com/causify-ai/helpers/issues/380


dev_scripts_helpers/notebooks/test/test_extract_notebook_images.py
```diff
@@ -0,0 +1,52 @@
+import logging
```
gpsaggese, line 1: I wouldn't check in the pngs.

We can assume that if there is a file, the file is correct, since we assume that nbconvert is correct

gpsaggese (commented):
Nit and merge.

gpsaggese:
Looks good. Merging

PR #426, opened 2025-03-25
CmTask11401_Find_a_way_to_keep_GitHub_repos_in_sync

heanhsok:
Task https://github.com/causify-ai/cmamp/issues/11401

- Sync GH issues labels


dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,318 @@
...
+
+class Label:
+
+    def __init__(self, name: str, description: str, color: str):
+        """
+        Initialize the label with name, description, and color.
+
+        :param name: Label name
+        :param description: Label description
+        :param color: Label color in hex format
```
sonniki, line 45: Here and everywhere: as per our conventions, param descriptions should start with a lowercase letter (except when it's a name like "GitHub")

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,318 @@
...
+            "Accept": "application/vnd.github.v3+json",
+        }
+
+    def get_labels(self, owner: str, repo: str) -> List[Label]:
+        """
+        Get all labels from a repository.
+
+        :param owner: GitHub repository owner/organization
+        :param repo: GitHub repository name
+        :return: List of Label objects
```
sonniki, line 81: No need to specify "List of" -- this is already clear from the typing hint

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,318 @@
...
+        response = requests.delete(url, headers=self.headers)
+        response.raise_for_status()
+        _LOG.info("Label deleted: %s", name)
+
+
+def load_manifest_to_labels(path: str) -> List[Label]:
+    """
+    Load label configurations from YAML file.
+
+    :param path: path to YAML file
```
sonniki, line 165: Instead of "YAML file" in all the function docstrings I would say "file with the label inventory". Just for the sake of focusing on the "semantics" of what we are referring to instead of the technical "implementation" details.

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,318 @@
...
+        _LOG.error("Error saving YAML file: %s", str(e))
+        sys.exit(1)
+
+
+def sync_labels(
+    client: GitHubClient, owner: str, repo: str, labels: List[Label], prune: bool
+) -> None:
+    """
+    Synchronize labels between config and repository.
+    """
```
sonniki, line 212: Missing param descriptions

dev_scripts_helpers/github/labels/gh_issues_labels.yml
```diff
@@ -0,0 +1,68 @@
+# Priority Labels.
```
sonniki, line 1: Does this mean that if one creates/updates/deletes a label now, they should also go to this file and update it? And then run the script to sync? Should we put this in the docs?

sonniki (changes requested):
All very nice! Just some style-related nitpicking from me.

dev_scripts_helpers/github/labels/gh_issues_labels.yml
heanhsok: Yes we need to document this process as well.

What I'm not sure yet are
- once the file is updated, who should do it and to which repos (since not everyone can see and has the permission to all the repos)?
- should we prune existing labels (there's an optional flag to do that) that are not in our manifest file?
- this python script is designed to run on one repo, I'm thinking of creating a shell script that would list all the repo (e.g. `gh repo list causify-ai`) and then execute this python script for each of the repo? (but this would also require the account that is used to execute the script to be able to see and update those settings)

WDYT?

heanhsok: > this python script is designed to run on one repo, I'm thinking of creating a shell script that would list all the repo (e.g. gh repo list causify-ai) and then execute this python script for each of the repo? (but this would also require the account that is used to execute the script to be able to see and update those settings)

Maybe we can also have a GH pipeline that detect changes in this file and then execute the changes when label file is merged to master?

This way, we are still relying on our script to perform the change and use GH Action as a way to detect and execute our python script.

sonniki: > once the file is updated, who should do it and to which repos (since not everyone can see and has the permission to all the repos)?
> this python script is designed to run on one repo, I'm thinking of creating a shell script that would list all the repo (e.g. gh repo list causify-ai) and then execute this python script for each of the repo? (but this would also require the account that is used to execute the script to be able to see and update those settings)

These questions are related IMO. I like the proposed approach of a shell script that would span all repos. I agree that this means that we have to have a user with a super access. @gpsaggese is that possible?

> should we prune existing labels (there's an optional flag to do that) that are not in our manifest file?

I would keep pruning off by default. We might have a situation when a user adds a label in a repo but forgets to update the manifest file, and we don't want to then delete this label once somebody runs the sync script. We could throw a logging message that there is label X in repo Y that is not reflected in the label inventory file.

sonniki: > Maybe we can also have a GH pipeline that detect changes in this file when label file is merged to master and then execute the changes?

I see the advantage of the user then only needing to update the yaml file when they update the labels. However, with running the script yourself there is more control, one can see logging messages, etc. So I would lean towards keeping the script in the hands of a human.

dev_scripts_helpers/github/labels/gh_issues_labels.yml
```diff
@@ -0,0 +1,68 @@
...
+# Issue Status.
+- name: Paused
+  description: An issue was started and then stopped
+  color: "#cccccc"
+- name: To close
+  description: An issue can be potentially closed
+  color: "#e6e6e6"
+# Contribution Types.
+- name: good first issue
+  description: Good for newcomers
```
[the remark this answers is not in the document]
sonniki: You don't agree with the suggestions for the yaml file? I see you've resolved the conversations but made no updates. It's okay, I just wanted to double check nothing slipped through the cracks

sonniki (commented):
See the question above regarding my previous comments on the yaml file. Otherwise, LGTM. 
Do you want to file follow-ups for the shell script and the docs? Or feel free to do it in this PR, up to you.

heanhsok:
> Do you want to file follow-ups for the shell script and the docs? Or feel free to do it in this PR, up to you.

Filed the followings as follow-ups so it's easier to review and discuss
- https://github.com/causify-ai/helpers/issues/438
- https://github.com/causify-ai/helpers/issues/439

sonniki (approved):
LGTM

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,326 @@
...
+        :param description: label description
+        :param color: label color in hex format
+        """
+        self.name = name
+        self.description = description
+        # Remove # prefix if present.
+        self.color = color.lstrip("#")
+
+    def __repr__(self):
+        return f"Label(name='{self.name}', description='{self.description}', color='{self.color}')"
```
gpsaggese, line 53: -> label 

symmetry is > than English :-)

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,326 @@
...
+        """
+        Initialize the label with name, description, and color.
+
+        :param name: label name
+        :param description: label description
+        :param color: label color in hex format
+        """
+        self.name = name
+        self.description = description
+        # Remove # prefix if present.
```
gpsaggese, line 49: -> `# comment`

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,326 @@
...
+    def __repr__(self):
+        return f"Label(name='{self.name}', description='{self.description}', color='{self.color}')"
+
+
+# #############################################################################
+# GitHubClient
+# #############################################################################
+
+
+class GitHubClient:
```
gpsaggese, line 61: Let's move this to helpers/hgithub.py

@tkpratardan is also working on related things so you might want to coordinate

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,326 @@
...
+
+        :param token: GitHub personal access token
+        """
+        self.token = token
+        self.headers = {
+            "Authorization": f"token {token}",
+            "Accept": "application/vnd.github.v3+json",
+        }
+
+    def get_labels(self, owner: str, repo: str) -> List[Label]:
```
gpsaggese, line 75: We could move owner and repo to the ctor and say "this is a class to operate on a repo", which is the typical use case.

Another approach is to derive a class `LabelProcessor` from `GitHubClient` or better use GitHubClient as a passed element. Composition is always > inheritance

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,326 @@
...
+
+        :param owner: GitHub repository owner/organization
+        :param repo: GitHub repository name
+        :return: label objects
+        """
+        url = f"https://api.github.com/repos/{owner}/{repo}/labels"
+        labels = []
+        page = 1
+        per_page = 50
+        while True:
```
gpsaggese, line 87: More comments explaining the chunks of work pls!

Also more _LOG.debug. The idea is that adding -v DEBUG helps with meaningful logging

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,326 @@
...
+                    name=item["name"],
+                    description=item["description"],
+                    color=item["color"],
+                )
+                for item in yaml_data
+            ]
+            return labels
+    except Exception as e:
+        _LOG.error("Error loading label inventory manifest file: %s", str(e))
+        sys.exit(1)
```
gpsaggese, line 182: I'd just rethrow here. Exiting is too C-language ;-)

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,326 @@
...
+    """
+    Save labels to the label inventory manifest file.
+
+    :param labels: label objects
+    :param path: path to save the label inventory manifest file to
+    """
+    try:
+        with open(path, "w") as file:
+            yaml_data = [
+                {
```
gpsaggese, line 195: Arch nit: this could be a `to_dict()` method, since it's general enough and avoids you to reach into the object (which is always a violation of encapsulation). In fact, I personally tend to make all variables private (like in C++)

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,326 @@
...
+                    "name": label.name,
+                    "description": label.description,
+                    "color": f"#{label.color}",
+                }
+                for label in labels
+            ]
+            yaml.dump(yaml_data, file, default_flow_style=False, sort_keys=False)
```
gpsaggese, line 202: We should comment the magic values, e.g., `default_flow_style`. Why are we setting to false?

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,326 @@
...
+    # Delete labels if pruning is enabled.
+    if prune:
+        for current_label in current_labels:
+            if current_label.name not in label_map:
+                try:
+                    client.delete_label(owner, repo, current_label.name)
+                except Exception as e:
+                    _LOG.error(
+                        "Error deleting label %s: %s", current_label.name, e
+                    )
```
gpsaggese, line 234: My (or ours) style is for objects to throw exceptions (typically dasserts) that have meaningful messages.
In this case client.delete_label does the work and if it can't, dasserts and the whole thing stops.

Catching and rethrowing is useful only if you can actually do something (rarely possible!) or you can add more context to the error.
In this case neither is possible so we should not catch and rethrow.

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,326 @@
...
+                current_label.description != label.description
+                or current_label.color != label.color
+            ):
+                # Label exists but needs update.
+                client.update_label(owner, repo, label)
+            else:
+                # Label exists and is identical.
+                _LOG.info("Label not changed: %s", label)
+        except Exception as e:
+            _LOG.error("Error processing label %s: %s", label.name, e)
```
gpsaggese, line 252: Same thing here. This is actually making the error worse, since it's better to get info about the failing point.

Long story short, we use try-except very rarely since it just makes the code more complicated.

gpsaggese (commented):
Adding @Shaunak01 @Prahar08modi @tkpratardan since they are the new blood.

The code is perfect, but there are some interesting architectural and OOP rules.

I would commit as a checkpoint and then file more bugs to:
1) clean up the code
2) unit test it
3) describe the logic and the principles

dev_scripts_helpers/github/sync_gh_issue_labels.py
heanhsok: As I'm refactoring this into its own class, I notice that we basically just re-inventing the wheel by rewriting what is already implemented in this package https://github.com/PyGithub/PyGithub (which is also already in our //helpers image). What if we scrape this entire class and just use the library? WDYT? @gpsaggese @sonniki

gpsaggese: Using the lib is always the way to go. Feel free to scrape the old code then.

heanhsok:
@gpsaggese @sonniki 
Wondering if there's any left to address in this PR, otherwise should we merge this?

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,244 @@
...
+import sys
+from typing import Dict, List
+
+import yaml
+
+try:
+    import github
+except ModuleNotFoundError:
+    subprocess.call(["sudo", "/venv/bin/pip", "install", "pygithub"])
+    import github
```
gpsaggese, line 28: Perfect candidate for Dockerized executable, no?
We can do it now or file a bug and have somebody else (an intern?) clean it up

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def _parse() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
+    )
+    hparser.add_verbosity_arg(parser)
+    parser.add_argument(
+        "--yaml",
+        "-y",
```
gpsaggese, line 137: let's avoid the short form since they make the command incomprehensible...

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,244 @@
...
+                name=label.name,
+                description=label.description if label.description else None,
+                color=label.color,
+            )
+        )
+    current_labels_map = {label.name: label for label in current_labels_data}
+    # Save the labels if backup is enabled.
+    if args.backup:
+        root_dir = hgit.get_client_root(False)
+        dst_dir = f"{root_dir}/dev_scripts_helpers/github/labels/backup"
```
gpsaggese, line 198: I would just do it here in a dir `tmp.backup_sync_gh_issue_labels.yaml`?

We use `tmp` for everything that doesn't need to be checked in

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,244 @@
...
+        file_name = f"labels.{args.owner}.{args.repo}.yaml"
+        file_path = f"{dst_dir}/{file_name}"
+        _save_labels(current_labels_data, file_path)
+        _LOG.info("Labels backed up to %s", file_path)
+    # Show the labels and exit if dry run is enabled.
+    if args.dry_run:
```
gpsaggese, line 204: This seems more of a "--print_labels".

I would imagine that `--dry_run` shows me what the script is going to do without doing it...

gpsaggese (commented):
Minor nits and merge.

dev_scripts_helpers/github/sync_gh_issue_labels.py
heanhsok: Changed the `dry_run` logic to print out the actions that would be taken without executing them

dev_scripts_helpers/github/sync_gh_issue_labels.py
heanhsok: We thought about doing it with through the use of invoke command. Not sure if we can combine the two approach?

heanhsok: Also we are not yet able to run tests of dockerized executable in the CI yet right?

gpsaggese: 1) Yes, I have a PR that will fix this.
You can file a follow up bug and add a TODO: Outsourceable

2) The import approach is problematic since if anyone import this (including pytest) stuff gets installed.

I would add an if False around that and ask people to install it manually before running

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,245 @@
...
+        :return: label as a dictionary
+        """
+        return {
+            "name": self._name,
+            "description": self._description,
+            "color": self._color,
+        }
+
+
+def _load_labels(path: str) -> List[Label]:
```
gpsaggese, line 87: FYI there is a school of thought that says that these should be static methods of Label, like `load_from_file` and `save_to_file`, since are part of the interface.

The design heuristics is that they are "verbs" applied to labels.

I personally like that approach, since it makes things tighter (e.g., everything is in a namespace).

It can also be a TODO: GSI for an intern

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    Load labels from label inventory manifest file.
+
+    :param path: path to label inventory manifest file
+    :return: label objects
+    """
+    try:
```
gpsaggese, line 94: I'm not super fond of this approach of try ... except

My reasoning is that everything can throw exceptions so what are we really achieving? I use try ... except when I expected something could go wrong and I want to handle it to help the caller, since I know the inside of the code and thus I can make a better decision.
Here the only info is that we print an error. But is it worth it the visual noise?
My sensibility says it's now worth it.

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def _main(parser: argparse.ArgumentParser) -> None:
+    args = parser.parse_args()
+    hdbg.init_logger(verbosity=args.log_level, use_exec_path=True)
+    # Load labels from label inventory manifest file.
+    labels = _load_labels(args.input_file)
+    labels_map = {label.name: label for label in labels}
+    token = os.environ.get(args.token_env_var, None)
+    hdbg.dassert(token)
```
gpsaggese, line 184: This is equivalent to `os.environ[args.token_env_var]`

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,245 @@
...
+    repo = client.get_repo(f"{args.owner}/{args.repo}")
+    # Get current labels from the repo.
+    current_labels = repo.get_labels()
+    current_labels_map = {label.name: label for label in current_labels}
+    # Execute code if not in dry run mode.
+    execute = not args.dry_run
+    # Save the labels if backup is enabled.
+    if args.backup:
+        root_dir = hgit.get_client_root(False)
+        dst_dir = f"{root_dir}/dev_scripts_helpers/github/labels/backup"
```
gpsaggese, line 196: I would just leave it in `.`. This is a temp file so we want to make the user's life easier, instead of having him search / use a complex path

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,245 @@
...
+        file_name = f"tmp.labels.{args.owner}.{args.repo}.yaml"
+        file_path = f"{dst_dir}/{file_name}"
+        if execute:
+            _save_labels(current_labels, file_path)
```
gpsaggese, line 200: else:
   _LOG.warning("Skipping saving labels as per user request")

My rule is when we are doing something "weird" we want to tell the user, so s/he knows what's happen

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,245 @@
...
+        _LOG.info("Labels backed up to %s", file_path)
```
gpsaggese, line 201: This should go in the if-then

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,245 @@
...
+    hsystem.query_yes_no(
+        "Are you sure you want to synchronize labels?", abort_on_no=True
+    )
+    # Delete labels if pruning is enabled.
+    if args.prune:
+        for current_label in current_labels:
+            if current_label.name not in labels_map:
+                if execute:
+                    current_label.delete()
+                _LOG.info("Label deleted: %s", current_label.name)
```
gpsaggese, line 212: Should go in the if-then and you can do

else:
   _LOG.info("The label %s will be deleted without --dry-run"

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,245 @@
...
+    if args.backup:
+        root_dir = hgit.get_client_root(False)
+        dst_dir = f"{root_dir}/dev_scripts_helpers/github/labels/backup"
+        file_name = f"tmp.labels.{args.owner}.{args.repo}.yaml"
+        file_path = f"{dst_dir}/{file_name}"
+        if execute:
+            _save_labels(current_labels, file_path)
+        _LOG.info("Labels backed up to %s", file_path)
+    # Confirm label synchronization.
+    hsystem.query_yes_no(
```
gpsaggese, line 203: Personally I like to add also a `--force` or `--no-interactive` flag to avoid to be requested for confirmation. The problem is that users don't want always to lock the program.

dev_scripts_helpers/github/sync_gh_issue_labels.py
```diff
@@ -0,0 +1,245 @@
...
+        current_label = current_labels_map.get(label.name)
+        if current_label is None:
+            # Label doesn't exist, create it.
+            if execute:
+                repo.create_label(
+                    name=label.name,
+                    color=label.color,
+                    description=label.description,
+                )
+            _LOG.info("Label created: %s", label.name)
```
gpsaggese, line 225: For me these are all _LOG.debug

Unix philosophy https://en.wikipedia.org/wiki/Unix_philosophy recommends "avoid unnecessary output"

My approach is I print very little at INFO level, if a user wants to see what happens can use a switch to print stats or run at DEBUG leve.

In this case I can be ok with being more chatty.
The way I would use these tools is like:
- run with dry-run to see what will happen
- check if what happens is what I want
- then run in one-shot and just know if things were fine or not

Does it make sense?

gpsaggese (commented):
The code is perfectly correct. I'm just nit picking on the architectural side and on the UI side. 

You can either fix these now or merge as it is and file an outsourceable issue to apply the following feedback.

Also we should find a way to document the style we use to write these tools so that everyone can follow it and we don't get everything working in a different way.

dev_scripts_helpers/github/labels/gh_issues_labels.yml
```diff
@@ -0,0 +1,68 @@
...
+  description: Simple tasks for newcomers
+  color: "#7057ff"
+- name: good second issue
+  description: Slightly more demanding tasks for newcomers
+  color: '#5F5187'
+- name: Outsource
+  description: Anybody can do it
+  color: "#919ebe"
+- name: help wanted
+  description: Extra attention is needed
```
gpsaggese, line 67: Instead of "help wanted"

Let's add
- Bounty

dev_scripts_helpers/github/labels/gh_issues_labels.yml
```diff
@@ -0,0 +1,68 @@
...
+  color: "#006b75"
+# Issue Status.
+- name: Paused
+  description: An issue was started and then stopped
+  color: "#cccccc"
+- name: To close
+  description: An issue can be potentially closed
+  color: "#e6e6e6"
+# Contribution Types.
+- name: good first issue
```
gpsaggese, line 57: Let's make them all the same style (e.g., no _ but space, etc)
-> Good first issue

dev_scripts_helpers/github/sync_gh_issue_labels.py
heanhsok: Added TODO in the code

dev_scripts_helpers/github/sync_gh_issue_labels.py
heanhsok: Got it. The logging here is more like outputs for users to view the actions that have been taken. Thus, I'm leaving it as _LOG.info as you suggested.

heanhsok:
Merging as checkpoints

I'll file the follow-up issue and reference the code in the master branch

Issue #433, opened 2025-03-26
Test dockerized pandoc


Shaunak01:
Check if the code needs testing 

script: `dev_scripts_helpers/documentation/dockerized_pandoc.py`
code: `helpers/hdocker.py`

FYI @gpsaggese

Shaunak01:
https://github.com/causify-ai/helpers/pull/455

Shaunak01:
closing as merged: https://github.com/causify-ai/helpers/pull/455

PR #455, opened 2025-03-31
HelpersTask433_Test_dockerized_pandoc

Shaunak01:
Issue: https://github.com/causify-ai/helpers/issues/433

### DIFF
--- a/dev_scripts_helpers/documentation/test/test_dockerized_pandoc.py
+++ b/dev_scripts_helpers/documentation/test/test_dockerized_pandoc.py
@@ -1,0 +1,94 @@
+import os
+
+import helpers.hdocker as hdocker
+import helpers.hunit_test as hunitest
+
+
+# #############################################################################
+# Test_Pandoc_Cmd_Conversion
+# #############################################################################
+
+
+class Test_Pandoc_Cmd_Conversion(hunitest.TestCase):
+
+    def test1(self) -> None:
+        """
+        Test `convert_pandoc_cmd_to_arguments` to parse a pandoc command string
+        into a dictionary.
+        """
+        # Create a sample pandoc command string.
+        cmd = (
+            "pandoc sample.md -o output.md --data-dir data "
+            "--template default --extract-media media -- --verbose --extra"
+        )
+        # Call function to test.
+        result = hdocker.convert_pandoc_cmd_to_arguments(cmd)
+        expected = {
+            "input": "sample.md",
+            "output": "output.md",
+            "in_dir_params": {
+                "data-dir": "data",
+                "template": "default",
+                "extract-media": "media",
+            },
+            "cmd_opts": ["--verbose", "--extra"],
+        }
+        # Check output.
+        self.assertEqual(result, expected)
+
+    def test2(self) -> None:
+        """
+        Test `convert_pandoc_arguments_to_cmd` to build a command string from a
+        dictionary of parameters.
+        """
+        # Prepare inputs.
+        params = {
+            "input": "sample.md",
+            "output": "output.md",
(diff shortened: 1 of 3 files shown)


Shaunak01:
@gpsaggese The fast test are failing as building the pandoc container takes too much time. What should be the workareound for this. The tests are done as we discussed

dev_scripts_helpers/documentation/test/test_dockerized_pandoc.py
```diff
@@ -0,0 +1,99 @@
...
+            "output": "output.md",
+            "in_dir_params": {
+                "data-dir": "data",
+                "template": "default",
+                "extract-media": "media",
+            },
+            "cmd_opts": ["--verbose", "--extra"],
+        }
+        # Check output.
+        self.assertEqual(result, expected)
```
gpsaggese, line 40: We prefer the idiom

```
act = pprint.pformat(result)
exp = """
"""
self.assert_equal(act, exp)
```

Since this is easier to copy/paste and the diff is better.

Can you use this?

dev_scripts_helpers/documentation/test/test_dockerized_pandoc.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+# #############################################################################
+# Test_run_dockerized_pandoc
+# #############################################################################
+
+
+class Test_run_dockerized_pandoc(hunitest.TestCase):
+
+    @pytest.mark.timeout(10)
```
gpsaggese, line 75: Can you mark it as superslow? There is a tag.

You can grep the documentation to see why / how.

dev_scripts_helpers/documentation/test/test_dockerized_pandoc.py
```diff
@@ -0,0 +1,99 @@
...
+        hio.create_dir(output_dir, incremental=True)
+        input_file = os.path.join(input_dir, "input.md")
+        output_file = os.path.join(output_dir, "sample.html")
+        # Build the pandoc command string.
+        cmd = f"pandoc {input_file} -o {output_file} --to=html --toc"
+        # Call the function.
+        hdocker.run_dockerized_pandoc(
+            cmd,
+            container_type="pandoc_texlive",
+            force_rebuild=True,
```
gpsaggese, line 92: I would not force rebuild, since if it's there it's better to use it.

helpers/hdocker.py
```diff
@@ -729,6 +729,8 @@ def convert_pandoc_cmd_to_arguments(cmd: str) -> Dict[str, Any]:
     # Parse known arguments and capture the rest.
     args, unknown_args = parser.parse_known_args(cmd)
     _LOG.debug(hprint.to_str("args unknown_args"))
+    # Filter out the option terminator if present.
```
gpsaggese, line 732: Can you add an example of when this will happen?

The comment explains what's happening (which is clear from the code), but we should explain why this is needed.

Makes sense?

dev_scripts_helpers/documentation/test/test_dockerized_pandoc.py
```diff
@@ -0,0 +1,99 @@
...
+            use_sudo=False,
```
gpsaggese, line 93: ok, but if it's not specified I think there is a get_sudo() approach that might allow you to deduce it automatically

gpsaggese (commented):
Good work. Nits and merge

dev_scripts_helpers/documentation/test/test_dockerized_pandoc.py
Shaunak01: noted

Issue #301, opened 2025-02-19
Dot at the beginning of a dir name is removed


sonniki:
`linters/amp_fix_md_links.py` turns the following link:
```
[`sprint_iteration.yml`](/.github/workflows/sprint_iteration.yml)
```
into the following
```
[`sprint_iteration.yml`](/github/workflows/sprint_iteration.yml)
```

However, the dot at the beginning of a dir name should not be removed. Need to investigate and fix.

sonniki:
@allenmatt10 
Along with the fix, we need to add this use case to the unit tests for this script to confirm that the fix works.

gpsaggese:
Yep, our approach is:
1) write a unit test to show that the problem is there
2) fix the issue
3) get the unit test to pass

sonniki:
Done, closing

PR #467, opened 2025-04-01
HelpersTask301_dot_at_beginning_of_dir_is_removed

allenmatt10:
Related to #301 

Updated `linters/amp_fix_md_links.py` to fix the directory naming issue in #301 .

Added the same use case to the unit tests in `linters/test/test_amp_fix_md_links.py` to confirm the fix

### DIFF
--- a/linters/amp_fix_md_links.py
+++ b/linters/amp_fix_md_links.py
@@ -38,7 +38,7 @@ def _make_path_absolute(path: str) -> str:
     :param path: the original path
     :return: the absolute path
     """
-    # Converts path beginning with './', '../', '../../'.
+    # Transform paths that begin with './', '../', '../../'.
     abs_path = re.sub(r"^(\./|\.\./|\.\./\.\./)*", "", path)
     abs_path = "/" + abs_path.lstrip("/")
     return abs_path
--- a/linters/amp_fix_md_links.py
+++ b/linters/amp_fix_md_links.py
@@ -38,7 +38,7 @@ def _make_path_absolute(path: str) -> str:
     :param path: the original path
     :return: the absolute path
     """
-    # Transform paths that begin with './', '../', '../../'.
+    # Converts path beginning with './', '../', '../../'.
     abs_path = re.sub(r"^(\./|\.\./|\.\./\.\./)*", "", path)
     abs_path = "/" + abs_path.lstrip("/")
     return abs_path
--- a/linters/amp_fix_md_links.py
+++ b/linters/amp_fix_md_links.py
@@ -38,11 +38,9 @@ def _make_path_absolute(path: str) -> str:
     :param path: the original path
     :return: the absolute path
     """
-    if path.startswith("./"):
-        abs_path = path[2:]
-    else:
-        abs_path = path.lstrip("/")
-    abs_path = "/" + abs_path
+    # Converts path beginning with './', '../', '../../'.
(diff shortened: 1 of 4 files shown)


allenmatt10:
@sonniki I have updated the code and test cases.

linters/amp_fix_md_links.py
```diff
@@ -38,8 +38,9 @@ def _make_path_absolute(path: str) -> str:
     :param path: the original path
     :return: the absolute path
     """
-    abs_path = path.lstrip("./")
-    abs_path = "/" + abs_path
+    # Converts path beginning with './', '../', '../../'.
```
sonniki, line 41: Comments should begin with imperatives.

linters/test/test_amp_fix_md_links.py
```diff
@@ -101,6 +101,50 @@ def test3(self) -> None:
...
         self.check_string(output)
 
+    def test4(self) -> None:
+        """
+        Test file path to retain directory name beginning with a dot.
+        """
+        file_path = "/.github/workflows/sprint_iteration.yml"
+        expected_path = "/.github/workflows/sprint_iteration.yml"
+        # Run.
+        actual_path = lafimdli._make_path_absolute(file_path)
```
sonniki, line 111: Nit: we usually just say `actual` and `expected` (instead of `actual_path/name/value...`), for unification.

linters/test/test_amp_fix_md_links.py
```diff
@@ -101,6 +101,50 @@ def test3(self) -> None:
...
+        Test to make file path absolute.
+        """
+        file_path = "../../.github/workflows/sprint_iteration.yml"
+        expected_path = "/.github/workflows/sprint_iteration.yml"
+        # Run.
+        actual_path = lafimdli._make_path_absolute(file_path)
+        # Check.
+        self.assertEqual(actual_path, expected_path)
+
     def _get_txt_with_incorrect_links(self) -> str:
```
sonniki, line 148: Can you also add the full link with `/.github` (you can just copy it from the issue) to this monster list of links? So that we can verify that it works end-to-end too. Don't forget to update [golden outcomes](https://github.com/causify-ai/helpers/blob/master/docs/coding/all.run_unit_tests.how_to_guide.md#update-golden-outcomes) for the relevant tests that use it.

sonniki (changes requested):

linters/test/outcomes/Test_fix_links.test1/output/test.txt
```diff
@@ -1,17 +1,18 @@
 # linter warnings
 
 /app/linters/test/outcomes/Test_fix_links.test1/tmp.scratch/test.md:38: '/helpersssss/hhhhgit.py' does not exist
-/app/linters/test/outcomes/Test_fix_links.test1/tmp.scratch/test.md:74: 'import_check/example/output/basic.png' does not follow the format 'figs/test.md/XYZ'
+/app/linters/test/outcomes/Test_fix_links.test1/tmp.scratch/test.md:41: '/.github/workflows/sprint_iteration.yml' does not exist
```
sonniki, line 4: The only nit: I would actually use one of the workflows that exists there and is also likely to not be removed at any point, like `fast_tests.yml`. So that we don't get this pointless "does not exist" complaint in the test. Could you please quickly change? And then it's a merge.

Issue #460, opened 2025-03-31
Use is_markdown_line_separator where needed


sonniki:
In [`helpers/hmarkdown.py`](https://github.com/causify-ai/helpers/blob/master/helpers/hmarkdown.py), there are a couple of TODOs to use `is_markdown_line_separator()` (from the same file) instead of explicitly declaring different types of separators. Let's address these TODOs.

aangelo9:
Comments are fixed. Awaiting new issue to be assigned.

PR #475, opened 2025-04-01
HelpersTask460_Use_is_markdown_line_separator_where_needed

aangelo9:
Related to #460 

Implemented `# TODO(gp): Use is_markdown_line_separator` in:

1. `process_single_line_comment()`
2. `format_headers()`


helpers/hmarkdown.py
```diff
@@ -147,13 +153,7 @@ def process_single_line_comment(line: str) -> bool:
         _LOG.debug("  -> do_continue=True")
         return do_continue
     # Skip frame.
-    # TODO(gp): Use is_markdown_line_separator
-    if (
-        re.match(r"\#+ -----", line)
-        or re.match(r"\#+ \#\#\#\#\#", line)
-        or re.match(r"\#+ =====", line)
-        or re.match(r"\#+ \/\/\/\/\/", line)
-    ):
```
sonniki, line 156: Can we confirm that both here and below `is_markdown_line_separator()` does indeed cover all these cases? 

Let's add more tests for `is_markdown_line_separator()` with lines that would be matched by `re.match(r"\#+ -----", line)`, `re.match(r"\#+ \#\#\#\#\#", line)` and so on. The goal is to make sure that we are not going to miss anything by doing this replacement. E.g. it seems like there were fewer symbols required for a match here than in `is_markdown_line_separator()` so they are not fully equivalent.

sonniki (changes requested):
Also let's not forget to [mention the issue](https://github.com/causify-ai/helpers/blob/master/docs/coding/all.submit_code_for_review.how_to_guide.md#mention-the-issue) in the PR's starting post.

helpers/hmarkdown.py
aangelo9: I have changed the logic of `is_markdown_line_separator()` to satisfy both the original if condition and the original function condition. And added test cases that covers both conditions. 

This is resolved, unless the hard coded logic of the original conditions is needed.

helpers/test/test_hmarkdown.py
```diff
@@ -161,7 +161,7 @@ def test_mode_headers1(self) -> None:
 
 class Test_is_markdown_line_separator1(hunitest.TestCase):
```
sonniki, line 162: @gpsaggese could you please check that these tests represent the desired behavior of the function? E.g. do we want to match lines without `#` at the start (as in `test_valid_separator1()`), or without the space between `#` and other characters (as in `test_valid_separator5()`)

sonniki (changes requested):
Good coverage -- just need to check with @gpsaggese (the author of the original code) if that's what we really want from the function (see comment below).

gpsaggese (commented):
Nothing else besides @sonniki comments

helpers/hmarkdown.py
```diff
@@ -21,11 +27,16 @@
 
 
 def is_markdown_line_separator(line: str) -> bool:
+    """
+    Check if the given line is a Markdown separator.
+
+    :return: A boolean if the line is a separator
```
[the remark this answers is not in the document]
aangelo9: Comments fixed. Awaiting a new issue to be assigned.

helpers/hmarkdown.py
```diff
@@ -21,11 +27,17 @@
 
 
 def is_markdown_line_separator(line: str) -> bool:
+    """
+    Check if the given line is a Markdown separator.
+
+    :param line: the current line of text being processed.
```
sonniki, line 33: No `.` at the end of param/return descriptions

sonniki (changes requested):
One last thing and then we'll merge

sonniki (approved):
LGTM

Issue #465, opened 2025-03-31
Generalize increase_chapter


sonniki:
In `helpers/hmarkdown.py`, generalize `increase_chapter()` to also decrease the header level, and rename accordingly, addressing [these TODOs](https://github.com/causify-ai/helpers/blob/bff9cfc764cbad5bfd9f93c0b0e9614eb324fae5/helpers/hmarkdown.py#L544).

allenmatt10:
@sonniki In addition to generalizing the function, shall I modify and run additional tests that were previously done in #463 ?

sonniki:
> [@sonniki](https://github.com/sonniki) In addition to generalizing the function, shall I modify and run additional tests that were previously done in [#463](https://github.com/causify-ai/helpers/issues/463) ?

Yes, of course -- the existing tests should be updated and new tests should be added for the new functionality.

sonniki:
Done

PR #482, opened 2025-04-03
HelpersTask465_generalize_increase_chapter

allenmatt10:
Related to #465 

Generalized `increase_chapter()` in `helpers/hmarkdown.py` to decrease header levels.

Modified and added new test cases in `helpers/test/test_hmarkdown.py`.


sonniki:
@samarth9008 could you please do the first review?

helpers/hmarkdown.py
```diff
@@ -542,23 +542,36 @@ def format_headers(in_file_name: str, out_file_name: str, max_lev: int) -> None:
 # TODO(gp): Add tests.
 # TODO(gp): Generalize this to also decrease the header level
 # TODO(gp): -> modify_header_level
```
samarth9008, line 544: We should remove the TODO if they are resolved to avoid confusion for future.

helpers/hmarkdown.py
```diff
@@ -542,23 +542,36 @@ def format_headers(in_file_name: str, out_file_name: str, max_lev: int) -> None:
...
-def increase_chapter(in_file_name: str, out_file_name: str) -> None:
+def modify_header_level(
+    in_file_name: str, out_file_name: str, signal: int
+) -> None:
     """
-    Increase the level of chapters by one for text in stdin.
+    Increase or decrease the level of chapters by one for text in stdin.
```
samarth9008, line 549: chapters -> headings

helpers/hmarkdown.py
```diff
@@ -542,23 +542,36 @@ def format_headers(in_file_name: str, out_file_name: str, max_lev: int) -> None:
...
 
-    :param in_file_name: The name of the input file to read
+    The function reads the input file, adjusts the chapter header levels
+    by one step based on the provided signal, and writes the modified
+    content to the output file.
```
samarth9008, line 553: no need for this

helpers/hmarkdown.py
```diff
@@ -542,23 +542,36 @@ def format_headers(in_file_name: str, out_file_name: str, max_lev: int) -> None:
 # TODO(gp): Add tests.
 # TODO(gp): Generalize this to also decrease the header level
 # TODO(gp): -> modify_header_level
-def increase_chapter(in_file_name: str, out_file_name: str) -> None:
+def modify_header_level(
+    in_file_name: str, out_file_name: str, signal: int
```
samarth9008, line 546: Instead of signal better to use a bool variable such as `increase: bool = True` (default behavior), since that was how the code behaved previously. If increase is false, we decrease. This way we won't need to change the code at many places.

helpers/hmarkdown.py
```diff
@@ -542,23 +542,36 @@ def format_headers(in_file_name: str, out_file_name: str, max_lev: int) -> None:
...
+        (-1) of the header level.
     """
     txt = hparser.read_file(in_file_name)
     #
     txt_tmp = []
     for line in txt:
         # TODO(gp): Use the iterator.
         line = line.rstrip(r"\n")
-        for i in range(1, 5):
+        for i in range(1, 7):
```
samarth9008, line 567: Why did we change it from 5 to 7? I would keep `range(1,6)` bcz max `#` we use are 5. We should also handle and test the edge case where we can't decrease if `#` is 1 or increase its 5

samarth9008 (commented):
Mid-level review as the code requires some arch changes.

Modify/improve/add news tests if needed.

helpers/hmarkdown.py
allenmatt10: I changed it to 7 because Markdown headings range from 1 to 6. I will update the code and make the necessary changes.

samarth9008: lets keep the old range of `range(1,5)`

allenmatt10: In that case, for test cases decreasing headers with size 5 will stay the same. Keeping range(1,6) solves all possible cases. Shall I keep the old range or update it to `range(1,6)`?

allenmatt10:
I've addressed the fixes required. Also, I have set the range in the function which satisfies the following edge cases:

- Increase for headings with 5 `#` to stay the same.
- Decrease for headings with 1 `#` to stay the same.
- Decrease for headings with 5 `#` will reduce by 1.

Let me know if any changes are to be made.

sonniki (changes requested):

helpers/hmarkdown.py
```diff
@@ -539,26 +539,35 @@ def format_headers(in_file_name: str, out_file_name: str, max_lev: int) -> None:
     hparser.write_file(txt_tmp, out_file_name)
 
 
-# TODO(gp): Add tests.
-# TODO(gp): Generalize this to also decrease the header level
-# TODO(gp): -> modify_header_level
-def increase_chapter(in_file_name: str, out_file_name: str) -> None:
+def modify_header_level(
+    in_file_name: str, out_file_name: str, increase: bool = True
```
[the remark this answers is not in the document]
samarth9008: > IMO it's less clear to have increase=False indicating that the level should be decreased, since "not increasing" != "decreasing".

Better, sg

helpers/hmarkdown.py
```diff
@@ -539,26 +539,35 @@ def format_headers(in_file_name: str, out_file_name: str, max_lev: int) -> None:
...
     """
     txt = hparser.read_file(in_file_name)
     #
     txt_tmp = []
+    increase_level = 1
     for line in txt:
         # TODO(gp): Use the iterator.
         line = line.rstrip(r"\n")
-        for i in range(1, 5):
+        for i in range(1, 6):
```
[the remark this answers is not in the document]
sonniki: Feel free to propose a solution based on your own ideas of what would be best.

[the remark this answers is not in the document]
allenmatt10: @sonniki 
Proposed solutions:

- We can modify `is_markdown_line_separator()` to exclude headings with 5`#`. Currently,  the test cases for this function do not test for headings with 5 `#`. Also, It's function calls would not have any serious impact if we follow this solution, so this change should be safe.
- We can use the iterator which handles all possible edge cases, as implemented in the previous commit.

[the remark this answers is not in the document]
sonniki: Go for num.1

dev_scripts_helpers/documentation/transform_notes.py
```diff
@@ -74,7 +74,7 @@ def _main(parser: argparse.ArgumentParser) -> None:
     elif cmd == "format":
         hmarkdo.format_headers(in_file_name, out_file_name, max_lev)
     elif cmd == "increase":
-        hmarkdo.increase_chapter(in_file_name, out_file_name)
+        hmarkdo.modify_header_level(in_file_name, out_file_name, cmd)
```
sonniki, line 77: Funny coincidence that here `cmd` is also "increase" but I would be careful -- if this code here is later refactored and `cmd` stops being exactly "increase", we'll have a problem. Just indicating `mode="increase"` is safer.

helpers/hmarkdown.py
```diff
@@ -539,27 +539,38 @@ def format_headers(in_file_name: str, out_file_name: str, max_lev: int) -> None:
...
     """
-    Increase the level of chapters by one for text in stdin.
+    Increase or decrease the level of headings by one for text in stdin.
 
-    :param in_file_name: The name of the input file to read
-    :param out_file_name: The name of the output file to write the
+    :param in_file_name: the name of the input file to read
+    :param out_file_name: the name of the output file to write the
         modified text to
+    :param mode: a string indicating the increase or decrease of the
```
sonniki, line 549: We usually avoid referring to type of vars in docstrings (e.g. "string"), since it can be already seen from the typing hints.

sonniki (changes requested):
A couple of small things left

samarth9008 (commented):
LGTM

Issue #294, opened 2025-02-18
Linter removed filename from fig pointer


sonniki:
Linting `docs/work_organization/ck.github_projects_process.reference.md`, got this:

![Image](https://github.com/user-attachments/assets/7ad38058-3710-41cd-9cab-2a2059427230)

Must have something to do with `amp_fix_md_links.py`. Need to investigate and fix.

sonniki:
The flow is:
1) Add unit tests to reproduce the problem
2) Debug the code and identify the source of the problem
3) Fix the problem
4) Confirm with the tests from (1) that the problem is gone

sonniki:
Done

PR #490, opened 2025-04-03
HelpersTask294_Linter_removed_filename_from_fig_pointer

indrayudd:
Addresses #294


linters/test/test_amp_fix_md_links.py
```diff
@@ -101,6 +101,36 @@ def test3(self) -> None:
...
+
+- A view can provide a specific perspective on the issues in the project, e.g.,
+  filter certain issues by their attributes, visualize them in a board/list
+
+<img src="figs/ck.github_projects_process.reference_figs/image1.png"
+    style="width:6.5in;height:0.31944in" />
+
+- We use a base
+  [<u>template</u>](https://github.com/cryptokaizen/cmamp/projects?query=is%3Aopen+is%3Atemplate)
+  for all of our projects to ensure consistency between projects
```
sonniki, line 121: I would remove most of the text that is not the link so that it's immediately clear what the focus is. You can keep something short there like `Text`, or `<Placeholder text>`.

linters/test/test_amp_fix_md_links.py
```diff
@@ -101,6 +101,36 @@ def test3(self) -> None:
...
+        """
+        file_name = "test_excerpt.md"
+        file_path = self._write_input_file(input_content, file_name)
+        # Run.
+        _, updated_lines, out_warnings = lafimdli.fix_links(file_path)
+        # Check.
+        actual = "\n".join(updated_lines)
+        expected = """
+        - Each GH project window consists of tabs called "views"\n\n- A view can provide a specific perspective on the issues in the project, e.g.,\n  filter certain issues by their attributes, visualize them in a board/list\n\n<img src="figs/ck.github_projects_process.reference_figs/image1.png"\n    style="width:6.5in;height:0.31944in" />\n\n- We use a base\n  [<u>template</u>](https://github.com/cryptokaizen/cmamp/projects?query=is%3Aopen+is%3Atemplate)\n  for all of our projects to ensure consistency between projects\n
```
sonniki, line 130: This is not very readable at the moment. Maybe with most of the text besides the link removed it will be better. But in general we want the link to be immediately visible at the first glance -- this is the most important thing. And it might also improve readability if instead of "\n" we actually do line breaks.

linters/test/test_amp_fix_md_links.py
```diff
@@ -101,6 +101,36 @@ def test3(self) -> None:
...
+    style="width:6.5in;height:0.31944in" />
+
+- We use a base
+  [<u>template</u>](https://github.com/cryptokaizen/cmamp/projects?query=is%3Aopen+is%3Atemplate)
+  for all of our projects to ensure consistency between projects
+        """
+        file_name = "test_excerpt.md"
+        file_path = self._write_input_file(input_content, file_name)
+        # Run.
+        _, updated_lines, out_warnings = lafimdli.fix_links(file_path)
```
sonniki, line 126: Doesn't Linter complain about `out_warnings` not being used?

sonniki (changes requested):
Goog debugging, just need some adjustments to the way it's framed

linters/test/test_amp_fix_md_links.py
indrayudd: Yes it does: `linters/test/test_amp_fix_md_links.py:126: [W0612(unused-variable), Test_fix_links.test4] Unused variable 'out_warnings' [pylint]`

Do you want me to add that to the output of the test? I originally omitted it because I didn't use the out_warnings variable during debugging.

linters/test/test_amp_fix_md_links.py
indrayudd: Removed everything except the link

linters/test/test_amp_fix_md_links.py
indrayudd: Removing the text should make it more readable

linters/test/test_amp_fix_md_links.py
sonniki: No, in this case Linter is pointing out the deficiency in the code. So you should fix it by, indeed, replacing it with `_`. The general approach is to try to minimize Linter complaints when possible, especially when it doesn't require much work.

indrayudd: Oops, your reply loaded right now! Incorporating `out_warnings` and resolving.

indrayudd: My apologies for misreading the response. I had inadvertently done the correct thing. Resolving.

sonniki (approved):
LGTM

Issue #300, opened 2025-02-19
Check repo name before shortening links


sonniki:
In `linters/amp_fix_md_links.py`, we have the following section handling http links:
```
    if link.startswith("http"):
        if not any(
            x in link
            for x in ["://github.com/cryptokaizen", "://github.com/causify-ai"]
        ):
            # The link is to an external resource; update is not needed.
            return line, warnings
        if "blob/master" not in link:
            # The link is not to a file (but, for example, to an issue);
            # update is not needed.
            return line, warnings
        # Leave only the path to the file in the link.
        link = link.split("blob/master")[-1]
```
What it does is it turns, for example, `https://github.com/causify-ai/helpers/blob/master/linters/amp_fix_md_links.py` into `/linters/amp_fix_md_links.py`. However, there is a problem if the link is supposed to point to a file in another repo. For example, if a file in `cmamp` has this link, then after cutting off the first part, the link will be actually broken because there is no `/linters/amp_fix_md_links.py` in `cmamp`.

We need to update the logic and only cut off the first part of the link if the repo name it contains is the same as the local repo (the one where the file that is being linted is located).

sonniki:
@indrayudd for testing, you can pick some file in the `tutorials` repo and test with a link pointing to it. 
For determining in which repo we are, feel free to browse libs in the `helpers` dir -- there must be already functions that do that.

indrayudd:
@sonniki since the `linters/amp_fix_md_links.py` file is in the helpers repo, should i code the logic such that it checks if the repo in the link is specifically `helpers` or do you want it to be more generalised; wherein if this codebase is ported into any other repo, it automatically checks for that local repo as well?

In the latter case, I can use the `_is_repo` function in `helpers/hgit.py`, which is a private function. Is it okay for me to create an external function from it for the `linters/amp_fix_md_links.py` script to use?

sonniki:
> if this codebase is ported into any other repo, it automatically checks for that local repo as well?

Yes, this. Because Linter can also be run in other repos as well (and `linters/amp_fix_md_links.py` is a part of Linter), and so, let's say, if we run Linter in `tutorials` on a file that has a link to a file located in `helpers`, we similarly don't want it to be "cut off". Does it make sense?''

> In the latter case, I can use the _is_repo function in helpers/hgit.py, which is a private function. Is it okay for me to create an external function from it for the linters/amp_fix_md_links.py script to use?

[...]

PR #498, opened 2025-04-05
HelpersTask300_Check_repo_name_before_shortening_links

indrayudd:
Addresses #300

Reverted the linted link in `linters/test/outcomes/Test_check_md_link_format.test1/input/test.md` and `linters/test/outcomes/Test_check_md_link_format.test2/input/test.md` to ensure it's a valid input for the tests.


linters/amp_fix_md_links.py
```diff
@@ -134,7 +134,11 @@ def _check_md_link_format(
             # The link is not to a file (but, for example, to an issue);
             # update is not needed.
             return line, warnings
-        # Leave only the path to the file in the link.
+        link_repo_short_name = hgit.extract_github_repo_short_name(link)
+        if not hgit.is_repo(link_repo_short_name):
```
samarth9008, line 138: Instead of making a new function, better to directly call `hrecouti.get_repo_config().get_repo_short_name()` to get repo short name

linters/amp_fix_md_links.py
```diff
@@ -134,7 +134,11 @@ def _check_md_link_format(
             # The link is not to a file (but, for example, to an issue);
             # update is not needed.
             return line, warnings
-        # Leave only the path to the file in the link.
```
samarth9008, line 137: No need to remove this comment. It should move to line 142

helpers/hgit.py
```diff
@@ -628,6 +635,22 @@ def _parse_github_repo_name(repo_name: str) -> Tuple[str, str]:
     return host_name, repo_name
 
 
+def extract_github_repo_short_name(url: str) -> str:
```
samarth9008, line 638: -> extract_repo_name

@sonniki Not sure if this is the right location. I would move this function to `amp_fix_md_links.py` file

helpers/hgit.py
```diff
@@ -628,6 +635,22 @@ def _parse_github_repo_name(repo_name: str) -> Tuple[str, str]:
...
+    """
+    Extract the short name of a github URL from a github repository link.
+    E.g., `https://github.com/causify-ai/helpers/blob/master/linters/amp_fix_md_links.py` -> `helpers`
+
+    :param url: github URL to be parsed
+    :return: the short name string
```
samarth9008, line 644: the short name string -> repo name

samarth9008 (commented):
Mid-review

helpers/test/test_hgit.py
```diff
@@ -698,3 +767,63 @@ def test2(self) -> None:
...
+
+
+# #############################################################################
+# Test_extract_github_repo_short_name
+# #############################################################################
+
+
+class Test_extract_github_repo_short_name(hunitest.TestCase):
+    """
+    Test to see if a short name of a github repo is extracted from its URL.
```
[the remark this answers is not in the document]
indrayudd: Do you mean modify the docstring?

linters/amp_fix_md_links.py
indrayudd: `hrecouti.get_repo_config().get_repo_short_name()` does not seem to be compatible with extracting short names from a URL. Are you suggesting doing a `hrecouti.get_repo_config().get_repo_short_name() == link_repo_short_name` check instead of making a new `is_repo` function?

helpers/hgit.py
sonniki: The question is whether we think this function would be useful outside of linting Markdown files. Since it has a URL as input, I think it's probably rare that it would be used for something other than transforming links in Markdown. So I'd probably agree that it should be moved to `amp_fix_md_links.py`

linters/amp_fix_md_links.py
samarth9008: Yes, that is correct

linters/test/test_amp_fix_md_links.py
```diff
@@ -349,3 +349,54 @@ def test_make_path_absolute4(self) -> None:
...
+        """
+        Test if in-repo links are parsed correctly.
+        """
+        # Prepare inputs.
+        link_text = "Fix Markdown links"
+        link = "https://github.com/causify-ai/helpers/blob/master/linters/amp_fix_md_links.py"
+        line = "[Fix Markdown links](https://github.com/causify-ai/helpers/blob/master/linters/amp_fix_md_links.py)"
+        file_name = (
+            "linters/test/outcomes/Test_check_md_link_format.test1/input/test.md"
+        )
```
sonniki, line 371: Since we are only really testing 2 lines, IMO there is no need to create input files for them specifically, it's fine to just declare them in the tests (like it's done in the test functions in `Test_fix_links`). Also, I would just test both cases in a single test function.

linters/amp_fix_md_links.py
```diff
@@ -87,6 +88,22 @@ def _check_md_header_exists(
...
 
+def _extract_repo_name(url: str) -> str:
+    """
+    Extract repo name from the github file link.
+    E.g., `https://github.com/causify-ai/helpers/blob/master/linters/amp_fix_md_links.py` -> `helpers`
+
+    :param url: github URL to be parsed
+    :return: repo name
+    """
+    # Extract the URL upto the repo name; e.g., `https://github.com/causify-ai/helpers`
```
sonniki, line 99: All comments should end in `.`

linters/amp_fix_md_links.py
```diff
@@ -87,6 +88,22 @@ def _check_md_header_exists(
...
+    repo_url = url.split("blob/master")[0][:-1]
+    # Extract the {organization}/{repo} part of the link and finally return the {repo} bit
+    repo = hgit._parse_github_repo_name(repo_url)[1]
+    repo_short_name = repo.split("/")[-1]
```
sonniki, line 103: Wouldn't it be easier to just do `.split("causify-ai/")[-1]`?

linters/amp_fix_md_links.py
```diff
@@ -134,6 +151,14 @@ def _check_md_link_format(
             # The link is not to a file (but, for example, to an issue);
             # update is not needed.
             return line, warnings
+        link_repo_short_name = _extract_repo_name(link)
+        if (
+            hrecouti.get_repo_config().get_repo_short_name()
+            != link_repo_short_name
+        ):
+            # The link points to another repo;
+            # update is not needed.
```
sonniki, line 160: Nit: this comment can probably fit into a single line.

linters/test/outcomes/Test_fix_links.test1/tmp.final.actual.txt
```diff
@@ -0,0 +1,120 @@
+# linter warnings
```
sonniki, line 1: You've checked in a bunch of `tmp` files from testing

linters/test/test_amp_fix_md_links.py
```diff
@@ -372,3 +392,38 @@ def test_make_path_absolute4(self) -> None:
...
+
+
+class Test_extract_repo_name(hunitest.TestCase):
+    """
+    Test _extract_repo_name() function.
+    """
+
+    def test1(self) -> None:
+        """
+        Test _extract_repo_name() function.
```
sonniki, line 409: No need to duplicate the docstring from the class docstring. You can either remove it completely (which is acceptable for simple tests that don't require extra explanation), or make more concrete for the specific tests (e.g. `Check that tutorials repo name is extracted correctly`).

sonniki (changes requested):

linters/amp_fix_md_links.py
```diff
@@ -87,6 +88,22 @@ def _check_md_header_exists(
...
+def _extract_repo_name(url: str) -> str:
+    """
+    Extract repo name from the github file link.
+    E.g., `https://github.com/causify-ai/helpers/blob/master/linters/amp_fix_md_links.py` -> `helpers`
+
+    :param url: github URL to be parsed
+    :return: repo name
+    """
+    # Extract the URL upto the repo name; e.g., `https://github.com/causify-ai/helpers`
+    repo_url = url.split("blob/master")[0][:-1]
```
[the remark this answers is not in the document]
indrayudd: Because the last character is a forward slash, thought it would just be a cleaner extraction without. I can alternatively split by "/blob/master" instead.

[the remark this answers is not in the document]
sonniki: Oh yes, just go ahead and split by `/blob...`

linters/amp_fix_md_links.py
```diff
@@ -87,6 +88,22 @@ def _check_md_header_exists(
...
+    E.g., `https://github.com/causify-ai/helpers/blob/master/linters/amp_fix_md_links.py` -> `helpers`.
+
+    :param url: github URL to be parsed
+    :return: repo name
+    """
+    # Extract the URL upto the repo name; e.g., `https://github.com/causify-ai/helpers`.
+    repo_url = url.split("/blob/master")[0]
+    # Extract the {organization}/{repo} part of the link and finally return the {repo} bit.
+    repo = hgit._parse_github_repo_name(repo_url)[1]
+    repo_short_name = repo.split("causify-ai/")[-1]
```
sonniki, line 103: Sorry, I still don't get why we need `_parse_github_repo_name()` and why not just do 
```
repo_url = url.split("/blob/master")[0].split("/causify-ai/")[-1]
```
Are there any corner cases I'm not seeing?

sonniki (changes requested):
One last thing to clear up and then we'll merge

linters/amp_fix_md_links.py
indrayudd: No, you're right. I simplified the code.

sonniki (approved):

Issue #462, opened 2025-03-31
Add tests for format_headers


sonniki:
Addressing the [TODO](https://github.com/causify-ai/helpers/blob/bff9cfc764cbad5bfd9f93c0b0e9614eb324fae5/helpers/hmarkdown.py#L490) in `helpers.hmarkdown()` of adding unit tests for `format_headers()`

sonniki:
It's fine since you've already done some work but why are you on this issue? We assign people to issues ourselves for easier monitoring and planning, please don't self-assign during this stage of internship.

sandeepthalapanane:
> [@sandeepthalapanane](https://github.com/sandeepthalapanane) It's fine since you've already done some work but why are you on this issue? We assign people to issues ourselves for easier monitoring and planning, please don't self-assign during this stage of internship.

Thanks for the clarification. I’ll avoid self-assigning issues going forward and will wait for assignments as per the process. Do you want me to work on this issue or leave it?

sonniki:
> Thanks for the clarification. I’ll avoid self-assigning issues going forward and will wait for assignments as per the process. Do you want me to work on this issue or leave it?

You can work on this issue since you've already started. I've unassigned you from the other one for now.

PR #501, opened 2025-04-05
Helpers task462 add tests for format headers

sandeepthalapanane:
Added unit tests for `format_headers()` in `helpers.hmarkdown()` #462


helpers/test/test_hmarkdown.py
```diff
@@ -1037,3 +1037,155 @@ def test8(self) -> None:
...
+    def test4(self) -> None:
+        """
+        Test inputs where max_level is inferred from the file content.
+        """
+        # Prepare inputs.
+        scratch_dir = self.get_scratch_space()
+        read_file = os.path.join(scratch_dir, "read_file.txt")
+        input_text = [
+            "# Chapter 1",
+            "max_level=2",
```
sonniki, line 1146: We could change this to `max_level=1` to better showcase that the max_level is then applied correctly (the expectation would be that `Chapter 1` is wrapped with heading markers but not `Section 1.1`.

helpers/test/test_hmarkdown.py
```diff
@@ -1037,3 +1037,155 @@ def test8(self) -> None:
...
         actual = hio.from_file(write_file)
         self.assertEqual(actual, expected)
+
+
+# #############################################################################
+# Test_format_headers1
+# #############################################################################
+
+
+class Test_format_headers1(hunitest.TestCase):
```
sonniki, line 1047: Don't forget to remove the original TODO once you address it.

helpers/test/test_hmarkdown.py
```diff
@@ -1037,3 +1037,155 @@ def test8(self) -> None:
...
+        """
+        # Prepare inputs.
+        scratch_dir = self.get_scratch_space()
+        read_file = os.path.join(scratch_dir, "read_file.txt")
+        input_text = [
+            "# Chapter 1",
+            "section text",
+        ]
+        input_text = "\n".join(input_text)
+        hio.to_file(read_file, input_text)
```
sonniki, line 1061: There is a lot of repetition in these test functions (creating scratch dir, writing and reading files, etc). Could we factor the common parts out in a helper method in this class? To which `input_text` and `expected` would be passed as parameters

sonniki (changes requested):
- Pls always [put a reference to the original issue](https://github.com/causify-ai/helpers/blob/master/docs/coding/all.submit_code_for_review.how_to_guide.md#mention-the-issue) in the starting post of the PR, to ease navigation
- Make sure to merge master regularly

helpers/test/test_hmarkdown.py
```diff
@@ -1037,3 +1037,121 @@ def test8(self) -> None:
...
+    def _helper_process(
+        self, input_text: list[str], expected: list[str], max_lev: int
+    ) -> None:
+        """
+        Helper function to process the given text with a specified maximum
+        level and compare the result with the expected output.
+
+        :param input_text: The text to be processed
+        :param expected: The expected output after processing the text
+        :param max_lev: The maximum heading level to be formatted
```
sonniki, line 1145: param/return descriptions start with lowercase

helpers/test/test_hmarkdown.py
```diff
@@ -1037,3 +1037,121 @@ def test8(self) -> None:
...
+        :return: None
```
sonniki, line 1146: no return description if it's None (just skip)

helpers/test/test_hmarkdown.py
```diff
@@ -1037,3 +1037,121 @@ def test8(self) -> None:
...
+            "No headings",
+        ]
+        expected = [
+            "Only text",
+            "No headings",
+        ]
+        self._helper_process(input_text, expected, max_lev=3)
+
+    def _helper_process(
+        self, input_text: list[str], expected: list[str], max_lev: int
```
sonniki, line 1137: We use `List` instead of `list` in typing hints (`from typing import List`)

helpers/test/test_hmarkdown.py
```diff
@@ -1037,3 +1037,121 @@ def test8(self) -> None:
...
+    ) -> None:
+        """
+        Helper function to process the given text with a specified maximum
```
sonniki, line 1140: Docstrings first lines begin with imperatives. E.g. just drop "Helper function to" from here.

sonniki (approved):

Issue #508, opened 2025-04-07
Generate current test coverage manually


samarth9008:
We want to generate current test coverage of helpers repo manually

We have an invoke for that `run_coverage_report` and a GH action which is running on cmamp

https://github.com/causify-ai/cmamp/blob/master/.github/workflows/test_coverage.yml

- [ ] Understand how the invoke is working

What we want is to run the invoke locally on the current master with the below command(might need to modify the command)

```
(client_venv.helpers) samarthk@dev1:~/src/helpers1$ invoke run_coverage_report --target-dir="." --no-publish-html-on-s3
```

The invoke will fail if tests are failing. OK to skip those tests for now when running the invoke.

The deliverable is an coverage report stored on the shared dir so anyone can access it. No need to create any PR or anything

The rationale behind this issue is to find untested code and file issues about it. This way we can find more issues for people being tested.

Lets make this high priority.

FYI @gpsaggese @sonniki

samarth9008:
As mentioned in the specs, you can put in the shared dir `/data/shared`, create a folder and then I'll take care of it.

Shaunak01:
@samarth9008 Sorry to ask again but I dont hve write access in the shared folder

```
(client_venv.helpers) shaunakd@dev1:~/src/helpers1$ ls -ld /data/shared
drwxr-xr-x 8 ubuntu 988 6144 Mar 26 19:31 /data/shared
```

Shaunak01:
PR: https://github.com/causify-ai/helpers/pull/511

samarth9008:
@gpsaggese @sonniki 

@Shaunak01 

Has generated the report locally. Since loading from S3 takes a while, the steps to run it locally is described below


```
> git checkout HelpersTask508_Generate_current_test_coverage_manually
> cd htmlcov
> python3 -m http.server 8000
```

Access http://localhost:8000/ locally on the web browser to see the coverage report.

PR #511, opened 2025-04-08
HelpersTask508_Generate_current_test_coverage_manually

Shaunak01:
Issue: https://github.com/causify-ai/helpers/issues/508
Added `/htmlcov` in the branch

### DIFF
--- a/helpers/lib_tasks_lint.py
+++ b/helpers/lib_tasks_lint.py
@@ -167,7 +167,8 @@ def lint_detect_cycles(  # type: ignore
         + hlitauti._to_single_line_cmd(docker_cmd_opts)
     )
     # Execute command line.
-    cmd = hlitadoc._get_lint_docker_cmd(docker_cmd_, stage, version)
+    base_image = ""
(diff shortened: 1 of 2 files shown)


samarth9008:
Uff so many files.

I was hoping it to be some kind of report file or something. Could we figure how to generate report that someone can easily access. With some kind of a link or anything

Shaunak01:
@samarth9008 
There is a way
So we can open the `/htmlcov/__init___py.html` on browser to view the report. That is the summary of everything.

- Open It Locally (One-Off)

```
cd htmlcov
python3 -m http.server 8000
```

Then open in browser:
```
http://localhost:8000
```
Others on the same network can access using others IP:
```
http://<your-ip>:8000
```

OR

We can print the page and share the PDF

Shaunak01:
Yup it was just for @samarth9008 to see the coverage output. I only have read access to 'data/' thus figured the browser way. Can we close @samarth9008

Shaunak01:
closing as done

Issue #487, opened 2025-04-03
Add Git actions for Codecov


Shaunak01:
We want to integrate Codecov to track and display coverage for our test suite. This will help us keep track of code quality and maintain a healthy testing culture. Below is the plan:

Steps and info for activating CodeCov: https://app.codecov.io/github/causify-ai/helpers/new

Steps:
- [x] Add codecove_requirements.txt and run pytest coverage action

Requirements              
```
coverage
codecov
```
In a GitHub Action, run tests and generate a coverage report:
```
pytest --cov --cov-branch --cov-report=xml
```

- [x] Select an upload token to add as a secret on GitHub
- [x] Add Codecov to GitHub Actions workflow yaml file
```
- name: Upload coverage reports to Codecov
    uses: codecov/codecov-action@v5
    with:
      token: ${{ secrets.CODECOV_TOKEN }}
```

@samarth9008 @dremdem as you are working in restructuring our git actions can you guide as to where we should add codecov actions? The token is already added to `helpers` so we just have to add the actions.

This is just for test purposes, we have not bought this service yet. I will document this once tested or we solidify the plan.

FYI @gpsaggese pls guide if you can as well.

Shaunak01:
closing with https://github.com/causify-ai/helpers/pull/506

samarth9008:
@dremdem 

do you mind taking a lead and mentor @Shaunak01 on this issue?

dremdem:
> where we should add codecov actions?

Let's not mix workflow responsibilities - let's keep this as a separate workflow.
We already have the 
https://github.com/causify-ai/cmamp/actions/workflows/test_coverage.yml
I'd suggest modifying that one.

Shaunak01:
> If there is no reason, feel free to transfer info to the old one and close this issue
We wanted to test in helpers only so opening the issue here from cmamp.

dremdem:
> Sure, can we start by modifying the helpers repo only

Yes, feel free to use [test_coverage.yml](https://github.com/causify-ai/cmamp/actions/workflows/test_coverage.yml) as a base.
First, I’d take a look at that GitHub Actions workflow and  invoke action that is calling the fast_tests with the [coverage](https://github.com/causify-ai/helpers/blob/b8c3b580d7d260b78d2c9d43e6ddde0f7e72506e/helpers/lib_tasks_pytest.py#L454) parameter.
Then, develop the new GitHub Actions workflow in the `helpers` repo.

[...]

PR #516, opened 2025-04-08
HelpersTask487_Add_Git_actions_for_Codecov_3

Shaunak01:
https://github.com/causify-ai/helpers/issues/487


Shaunak01:
@gpsaggese @dremdem 

Take a look at how the coverage looks: https://app.codecov.io/github/causify-ai/helpers/tree/HelpersTask487_Add_Git_actions_for_Codecov_3/

We get a 14 day free trial for complete org so we can do that if we like this service.

.github/gh_requirements.txt
```diff
@@ -2,3 +2,4 @@ invoke
 tqdm
 s3fs
 boto3 >= 1.20.17
+coverage
```
gpsaggese, line 5: Alphabetical order. Can you pls add a comment in the file?

.github/workflows/test_coverage.yml
```diff
@@ -0,0 +1,78 @@
+name: Test coverage
```
gpsaggese, line 1: Pls check with @samarth9008 and @dremdem how to make this general. 

Not sure if we want to have a coverage for fast, slow, super slow tests. Is it possible to have 3 views?
Ideally we would love to compute them separately and then all together like "total coverage" once a week.

Can you file a doc tracking the design of this?

.github/gh_requirements.txt
```diff
@@ -2,3 +2,4 @@ invoke
 tqdm
```
gpsaggese, line 2: Ok with having a single gh_req even if we don't need coverage all the times.
Unfortunately there is no easy way to compose gh_reqs, or can you do `pip install -r ... -r gh_req.coverage.txt`?

helpers/test/test_lib_tasks_gh.py
```diff
@@ -34,9 +35,8 @@ def test_get_gh_issue_title1(self) -> None:
         self.assert_equal(str(act), str(exp))
 
     @pytest.mark.skipif(
-        not hgit.is_in_helpers_as_supermodule(),
-        reason="""Skip unless helpers is the supermodule. Fails when updating submodules;
-            passes in fast tests super-repo run. See CmTask10845.""",
+        not hgit.is_in_helpers_as_supermodule() or os.getenv("CSFY_CI") == "true",
```
gpsaggese, line 38: there is a is_ci function somewhere in hserver

samarth9008:
@dremdem 

Let me know if you have any blockages. 

@Shaunak01 

The working should be similar to the current coverage run

.github/workflows/test_coverage.yml
```diff
@@ -0,0 +1,78 @@
...
+          CSFY_ECR_BASE_PATH: ghcr.io/${{ github.repository_owner }}
+          CSFY_AWS_S3_BUCKET: ${{ vars.CSFY_AWS_S3_BUCKET }}
+        run: invoke run_coverage_report .
+
+      # Optionally add a debugging step to verify that coverage.xml was generated:
+      - name: List workspace contents after coverage generation
+        run: ls -alh ${{ github.workspace }}
+
+      - name: Upload coverage to Codecov
+        if: always()
```
dremdem, line 71: I'd add an explanatory comment above.

.github/workflows/test_coverage.yml
```diff
@@ -0,0 +1,78 @@
...
+        uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: ./coverage.xml
+          flags: helpers
+          name: helpers-test-coverage
+          fail_ci_if_error: true
```
dremdem, line 78: Just for my understanding: what does this parameter do?

.github/workflows/test_coverage.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Run tests and generate coverage XML inside Docker
+        env:
+          GH_ACTION_ACCESS_TOKEN: ${{ secrets.GH_ACTION_ACCESS_TOKEN }}
+          CSFY_AWS_ACCESS_KEY_ID: ${{ env.AWS_ACCESS_KEY_ID }}
+          CSFY_AWS_SECRET_ACCESS_KEY: ${{ env.AWS_SECRET_ACCESS_KEY }}
+          CSFY_AWS_SESSION_TOKEN: ${{ env.AWS_SESSION_TOKEN }}
+          CSFY_AWS_DEFAULT_REGION: ${{ env.AWS_DEFAULT_REGION }}
+          CSFY_ECR_BASE_PATH: ghcr.io/${{ github.repository_owner }}
+          CSFY_AWS_S3_BUCKET: ${{ vars.CSFY_AWS_S3_BUCKET }}
+        run: invoke run_coverage_report .
```
dremdem, line 64: Am I right in saying that the `run_coverage_report` doesn't include the `superslow` tests?

helpers/lib_tasks_pytest.py
```diff
@@ -864,6 +864,8 @@ def run_coverage_report(  # type: ignore
         if exclude_from_report is not None:
             report_html_cmd += f" --omit={exclude_from_report}"
         report_cmd.append(report_html_cmd)
+    # Generate an XML report for Codecov.
```
dremdem, line 867: Probably it worth to add the comment into the docstring.

helpers/test/test_lib_tasks_gh.py
dremdem, line 38: What are we trying to fix here?

dremdem (commented):
For draft PR is looking good. 
Small nits. 
Considering
https://github.com/causify-ai/helpers/pull/516/files#r2036238705
Need to rework it as separate GH actions workflows for fast, slow and superslow.

.github/workflows/test_coverage.yml
Shaunak01: That is correct. No superslow

Shaunak01:
@dremdem 

I am trying to generate separate and combined coverage reports for fast and slow tests to see sep reports on copdecov. Initially, the default approach worked by using the implicit normalization of file paths in the default `.coverage` file, but in my approach we renamed the files and explicitly set the data file with the `COVERAGE_FILE` variable, absolute paths (like `/app/config_root/__init__.py` and `/app/conftest.py`) began causing errors because Coverage couldn't find the corresponding local source. Now, I am exploring ways to either map these paths, ignore errors, or otherwise replicate the default behavior so that every file is included in the report without manual omission.

helpers/test/test_lib_tasks_gh.py
Shaunak01: The test fails in CI for some reasons. There are many tests rn which only pass locally

Shaunak01:
@gpsaggese 

Pls take a look at: https://app.codecov.io/github/causify-ai/helpers

We can have the coverage for just the fast tests, slow tests or you can select both flags to see combined

FYI the combined and helpers flag are useless so ignore

.github/workflows/test_coverage.yml
Shaunak01: `fail_ci_if_error: true` causes the job to fail if the upload to Codecov fails, helping to catch issues with the coverage reporting.

.github/workflows/test_coverage.yml
Shaunak01: yes

dremdem:
> @gpsaggese @dremdem
> 
> Pls take a look at: https://app.codecov.io/github/causify-ai/helpers
> 
> We can have the coverage for just the fast tests, slow tests or you can select both flags to see combined
> 
> FYI the combined and helpers flag are useless so ignore

It looks good. 
Probably later we can also integrate it into PRs statuses: 
https://github.com/marketplace/codecov

.github/workflows/test_coverage.yml
```diff
@@ -0,0 +1,120 @@
...
+          CSFY_AWS_SESSION_TOKEN: ${{ env.AWS_SESSION_TOKEN }}
+          CSFY_AWS_DEFAULT_REGION: ${{ env.AWS_DEFAULT_REGION }}
+          CSFY_ECR_BASE_PATH: ghcr.io/${{ github.repository_owner }}
+          CSFY_AWS_S3_BUCKET: ${{ vars.CSFY_AWS_S3_BUCKET }}
+        run: invoke run_fast_coverage .
+
+      # After the test run, rename the generated coverage.xml to coverage_fast.xml.
+      - name: Rename Fast Test Coverage report
+        if: always()
+        run: mv ./coverage.xml ./coverage_fast.xml
```
dremdem, line 74: Wondering, why we need it if we will specify the `flags: fast` in the next step?

.github/workflows/test_coverage.yml
```diff
@@ -0,0 +1,120 @@
...
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: ./coverage_fast.xml
+          flags: fast
+          name: fast-test-coverage
+          # `fail_ci_if_error: true` instructs the action to mark the CI as failed
+          # if there is any error during the upload to Codecov.
+          fail_ci_if_error: true
+
+      # This step is used to trigger the slow test coverage generation using the invoke task.
+      - name: Run Slow test and generate report
```
dremdem, line 93: Probably, we need separate workflows for the fast, slow and super slow tests.

.github/workflows/test_coverage.yml
Shaunak01: So if you look at the invoke function in `helpers/lib_tasks_pytest.py`. Initially we had combined coverage only now i have introduced separate fast and slow coverage runs. But both generate the same file `coverage.xml` due to internal logic of porting. If i rename in the internal logic, we have to debug some porting logic as codecov is not picking up other files. So a better solution is

Generate Fast test coverage - `coverage.xml` -> rename 'coverage_fast.xml'-> upload
Generate Slow test coverage - `coverage.xml` -> rename 'coverage_slow.xml'-> upload

Codecov combines so we dont need to.

dremdem: I'm not sure I got the point. To be honest, I'm lost.
@Shaunak01 Let's chat tomorrow.

Shaunak01: Works!

helpers/lib_tasks_pytest.py
```diff
@@ -874,6 +877,136 @@ def run_coverage_report(  # type: ignore
...
 
+def _get_inclusion_settings(target_dir: str) -> Tuple[str, Optional[str]]:
+    """
+    Helper function to determine the directories to include and exclude in the
+    coverage report based on the target directory.
+
+    :param target_dir: directory for coverage stats. Use "." to indicate all directories
+    :return: tuple (include_in_report, exclude_from_report) where:
+             - include_in_report is a glob pattern for files to include
+             - exclude_from_report is a comma-separated glob pattern for files to exclude (or None if not applicable)
```
gpsaggese, line 888: Reflow

helpers/lib_tasks_pytest.py
```diff
@@ -864,6 +865,8 @@ def run_coverage_report(  # type: ignore
         if exclude_from_report is not None:
             report_html_cmd += f" --omit={exclude_from_report}"
         report_cmd.append(report_html_cmd)
+    # Generate an XML report for Codecov.
+    report_cmd.append("coverage xml -o coverage.xml")
```
gpsaggese, line 869: Can we disable these comments

image

 from Codecov? 

I think they are cool, but for now they add a lot of noise.

Shaunak01:
closing with follow up PR at: https://github.com/causify-ai/helpers/pull/598

Issue #334, opened 2025-03-03
Avoid empty lines in code


sonniki:
We want to avoid empty lines within functions and methods. E.g., in
```
def func(x):
    y = x + 1
    
    print(y)
```
the empty line needs to be either removed:
```
def func(x):
    y = x + 1
    print(y)
```
or filled in with a comment:
```
def func(x):
    y = x + 1
    # Print the result.
    print(y)
```
An acceptable (but undesirable) filler is an empty comment: 
```
def func(x):
    y = x + 1
    #
    print(y)
```

We should add a Linter step to 
- check for such empty lines (of course, it should ignore empty lines between the functions, etc)
- either (1) fix the problem, by (1.1) removing the empty line or (1.2) filling it with an empty comment
- or (2) issue a complaint so that the user can fix it themselves

The benefit of (1) is that we guarantee no empty lines in code as long as Linter is run. 
The benefit of (2) is that we give the user a chance to add a meaningful comment.
@gpsaggese what do you Linter should do here?

P.S. I would say, this is a good second/third issue for the interns once they have gotten their feet wet with something very simple first.

gpsaggese:
I would remove the lines for simplicity.

I'm experimenting with some LLM prompts to automatically add comments to chunks of codes that need it.
The goal is to get linte + LLMs to keep our code in nice shape, ofc a human should be in the loop. We can't let the machine run amok (yet)

sonniki:
@allenmatt10 [this doc](https://github.com/causify-ai/helpers/blob/master/docs/coding/all.developing_linter.how_to_guide.md) should tell you everything that you need for creating a new Linter step. As always, a good rule of thumb is looking at how the existing Linter scripts are structured and following their example.

gpsaggese:
Yes, let's do it in the linter without LLMs. After a few tries, LLMs seem to have problems understanding the concept of "empty line," in the same way, they are "not good" at math.
Also it's too expensive (and uncool) to pass all the code to the LLM for a simple operation that can be Python-ized.

@allenmatt10 pls make a proposal of how to solve the problem.

It's ok to make some simplifying assumptions, e.g.,
1) Add a switch to check for empty lines (and assert) and one to remove them
    - @sonniki we can use the check to make sure the code has been linted
2) Look for a function (starting with `def`) and just remove all the empty lines
3) IIRC the linter does a pass of `black` last adding empty lines when needed (e.g., in functions inside other functions)
[...]

PR #528, opened 2025-04-10
HelpersTask334_avoid_empty_lines_in_code

allenmatt10:
Related to #334

As per https://github.com/causify-ai/helpers/issues/334#issuecomment-2791115194, this PR reviews some use cases to make sure everyone's on the same page on the I/O behavior before its implementation.

Sample case 1:
```
import os
import numpy as np
import pandas as pd

from typing import List, Tuple

def func1(a:str)->str:
    """
    Print the string passed.

    :param a: String to be printed
    """

    print(a)

fun1("demo")

print("Outside function")
```
Output 1: Line removed (Docstring preserved)
```
import os
import numpy as np
import pandas as pd

from typing import List, Tuple

def func1(a:str)->str:
    """
    Print the string passed.

    :param a: String to be printed
    """
    print(a)

fun1("demo")

print("Outside function")
```
Sample case 2: Function nested inside classes.
```
import os
import numpy as np
import pandas as pd

from typing import List, Tuple

def func1(a:str)->str:
    """
    docstring 1

    :param a: description
    """
    print("Inside fun1")

    print(a)

class TestDemo():

    def func2(a: str) -> str:
        """
        docstring 2

        :param a: description
        : return: description
        """

        print("Inside fun2")

        return a


    def func3(a: str) -> str:
        """
        docstring 3

        :param a: description
        :return : description
        """
        print("Inside fun3")

        return a


fun1("demo")

print("Outside class")

```
Output 2:
```
import os
import numpy as np
import pandas as pd

from typing import List, Tuple

def func1(a:str)->str:
    """
    docstring 1

    :param a: description
    """
    print("Inside fun1")
    print(a)

class TestDemo():

    def func2(a: str) -> str:
        """
        docstring 2

        :param a: description
        : return: description
        """
        print("Inside fun2")
        return a


    def func3(a: str) -> str:
        """
        docstring 3

        :param a: description
        :return : description
        """
        print("Inside fun3")
        return a


fun1("demo")

print("Outside class")

```

Instead of removing the lines, it can be replaced by `#` as well. Let me know if these I/O behaviors are okay before I start implementing the linter action.


sonniki:
@allenmatt10 

The test cases look good, although ideally we would want to remove the empty line in here too:
```
fun1("demo")

print("Outside class")
``` 
But in practice, in our libraries we shouldn't have code that is not inside some function, so it's fine to ignore for now.
BTW "functions nested inside classes" are called methods:)

> Instead of removing the lines, it can be replaced by # as well.

We can add that as an option but when we call Linter we won't be able to pass an extra argument to control this behavior (technically we could but we definitely don't want to). So one of the options should be given the default status, and for now we've decided that it should be the removal of empty lines.

> Let me know if these I/O behaviors are okay before I start implementing the linter action.

I see you've already checked in some files. Do you want to work on them some more or is this PR ready for review?

allenmatt10:
> Do you want to work on them some more or is this PR ready for review?

@sonniki The PR is ready for review. Let me know if any changes are to be made, before I update `docs/coding/all.developing_linter.how_to_guide.md` to include this new linter script.

linters/amp_fix_empty_lines.py
```diff
@@ -0,0 +1,171 @@
...
+    return "\n".join(cleaned_file)
+
+
+def _clean_function_block(
+    func_lines: List[str], docstring_done: bool, docstring_open: bool
+) -> List[str]:
+    """
+    Remove empty lines inside function.
+
+    :param func_lines: list of lines in a function
```
sonniki, line 83: No need to refer to the type ("list")

linters/test/test_amp_fix_empty_lines.py
```diff
@@ -0,0 +1,331 @@
...
+        """
+        Test cleaning empty lines in a single function.
+        """
+        text = '''import os
+import numpy as np
+import pandas as pd
+
+from typing import List, Tuple
+
+def func1(a:str)->str:
```
sonniki, line 22: Even though it's not a "real function", let's still make it look neat, following style conventions of our normal functions.

linters/test/test_amp_fix_empty_lines.py
```diff
@@ -0,0 +1,331 @@
...
+    """
+    Print the string passed.
+
+    :param a: String to be printed
+    """
+    print(a)
```
sonniki, line 28: Since there is only one line in this function, is this test testing that nothing happens when there is a single line? We could indicate that in the test docstring

linters/test/test_amp_fix_empty_lines.py
```diff
@@ -0,0 +1,331 @@
...
+from typing import List, Tuple
+
+def func1(a:str)->str:
+    """
+    docstring 1
+
+    :param a: description
+    """
+    print("Inside fun1")
+    print(a)
```
sonniki, line 72: There are no empty lines here and in the methods below. Presumably, because you ran Linter on this file and it removed them... The solution would be to move the inputs to a separate file in the `input` dir. Since it will be a txt, Linter will not fix it. Also it will be better visually, right now the tests are not very readable.

linters/test/test_amp_fix_empty_lines.py
```diff
@@ -0,0 +1,331 @@
...
+
+def func1(a:str)->str:
+    """
+    Print the string passed.
+
+    :param a: String to be printed
+    """
+    print(a)
+
+fun1("demo")
```
sonniki, line 30: `func1`
Sorry, again, ofc it's not a real function and it doesn't really matter, it's just the matter of everything being tidy and clean

sonniki (changes requested):

linters/amp_fix_empty_lines.py
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env python
+"""
+Remove empty lines within a function.
+
+Import as:
+
+import linters.amp_fix_empty_lines as lafiemli
```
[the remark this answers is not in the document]
gpsaggese: Yes, let's do `amp_remove_empty_lines_from_python`

We are not using a Commodore64, which had 48KB of RAM. We don't need to save bytes.

linters/amp_fix_empty_lines.py
```diff
@@ -0,0 +1,171 @@
...
+from typing import List
+
+import helpers.hdbg as hdbg
+import helpers.hio as hio
+import helpers.hparser as hparser
+import linters.action as liaction
+import linters.utils as liutils
+
+
+def update_function_blocks(file_content: str) -> str:
```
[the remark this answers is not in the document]
gpsaggese: Sonya is right: clarity is what matters the most.

The best approach is to split the logic in clear steps.
You can use a variable to represent the state in which you are (e.g., looking_for_func, inside_func, etc).
Once you are inside a function just remove all empty lines.

gpsaggese (commented):
Our approach is to make quick baby steps.

In this case
- one PR to create unit tests with before and after
- one PR with the skeleton of the algo
- one PR with the actual algo making the unit tests
- one PR integrating it in the linter

linters/test/test_amp_fix_empty_lines.py
gpsaggese: Everything needs to be always tidy and clean. There is no other way of doing things.

allenmatt10: Thank you for your feedback. I will follow the procedures more carefully from now on.

helpers/hstring.py
```diff
@@ -94,6 +94,47 @@ def get_docstring_line_indices(lines: List[str]) -> List[int]:
...
+            start = i
+            i += 1
+            while i < n:
+                current_line = lines[i]
+                if current_line.strip() == "":
+                    # Skip empty line inside the function or method.
+                    i += 1
+                    continue
+                current_indent = len(current_line) - len(current_line.lstrip())
+                if current_indent <= base_indent:
```
sonniki, line 122: The logic about indentation used to determine the function end should be explained with comments.

helpers/hstring.py
```diff
@@ -94,6 +94,47 @@ def get_docstring_line_indices(lines: List[str]) -> List[int]:
     return docstring_line_indices
 
 
+def get_function_indices(lines: List[str]) -> List[int]:
```
sonniki, line 97: Please add a TODO to add tests (make sure to follow our conventions on formatting the TODO comment)

linters/test/test_amp_remove_empty_lines.py
```diff
@@ -0,0 +1,74 @@
...
+        test_input_dir = self.get_input_dir()
+        text_file_path = os.path.join(test_input_dir, "test.txt")
+        text = hio.from_file(text_file_path)
+        # Run.
+        actual = lareemli.update_function_blocks(text)
+        # Check.
+        test_output_dir = self.get_output_dir()
+        output_file_path = os.path.join(test_output_dir, "test.txt")
+        expected = hio.from_file(output_file_path)
+        self.assertEqual(expected, "\n".join(actual))
```
sonniki, line 28: Is this repeated fully in all the tests? In this case it should be factored out in a helper method, which will be called in each test.

helpers/hstring.py
```diff
@@ -94,6 +94,47 @@ def get_docstring_line_indices(lines: List[str]) -> List[int]:
...
+        line = lines[i]
+        match = re.match(r"(\s*)def\s+\w+", line)
+        if match:
+            base_indent = len(match.group(1))
+            start = i
+            i += 1
+            while i < n:
+                current_line = lines[i]
+                if current_line.strip() == "":
+                    # Skip empty line inside the function or method.
```
sonniki, line 118: "Skip" is a bit misleading here, since you are still adding this line to your list of lines within functions. Probably should be the opposite, e.g. "Register empty lines that are inside the function."

helpers/hstring.py
```diff
@@ -94,6 +94,47 @@ def get_docstring_line_indices(lines: List[str]) -> List[int]:
...
+                    i += 1
+                    continue
+                current_indent = len(current_line) - len(current_line.lstrip())
+                if current_indent <= base_indent:
+                    while True:
+                        # Empty lines after the function is included in the function
+                        # so it must be removed.
```
sonniki, line 125: "Do not register empty lines that follow the function."

sonniki (changes requested):
This is much better. Some more things

allenmatt10:
I’ve made the required changes, but the slow tests and linter checks are failing for this commit. On reviewing the logs, both show a common error:
```
docker: Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /home/runner/work/helpers/helpers
```

I'm assuming it could be related to a recent push (#544) which updated `/dev_scripts_helpers/documentation/lint_notes.py`

@sonniki Could this be related, in your opinion?

helpers/hstring.py
```diff
@@ -94,6 +94,48 @@ def get_docstring_line_indices(lines: List[str]) -> List[int]:
...
+    :param lines: the code lines to check
+    :return: the indices of function lines
+    """
+    function_line_indices = []
+    i = 0
+    n = len(lines)
+    while i < n:
+        # Process each line to find function header.
+        line = lines[i]
+        match = re.match(r"(\s*)def\s+\w+", line)
```
gpsaggese, line 111: When there are regex let's put a comment with an example of what we are looking for

helpers/hstring.py
```diff
@@ -94,6 +94,48 @@ def get_docstring_line_indices(lines: List[str]) -> List[int]:
     return docstring_line_indices
 
 
+# TODO(allenmatt10): Add tests.
+def get_function_indices(lines: List[str]) -> List[int]:
```
gpsaggese, line 98: lines -> txt

Let's say what it represents and now how it's encoded.
Can you change also the function below? (Which is not great)

gpsaggese, line 98: get_ -> find_

helpers/hstring.py
```diff
@@ -94,6 +94,48 @@ def get_docstring_line_indices(lines: List[str]) -> List[int]:
...
+    """
+    Get indices of lines of code that are inside functions and methods.
```
gpsaggese, line 100: let's add a bit of an example since we will forget what we are doing in 6 months

helpers/hstring.py
```diff
@@ -94,6 +94,48 @@ def get_docstring_line_indices(lines: List[str]) -> List[int]:
...
+
+    :param lines: the code lines to check
+    :return: the indices of function lines
+    """
+    function_line_indices = []
```
gpsaggese, line 105: Let's add some _LOG.debug to help debugging

Our rule is: "if there is a problem, one should enable -v DEBUG and get enough info to debug"

helpers/hstring.py
```diff
@@ -94,6 +94,48 @@ def get_docstring_line_indices(lines: List[str]) -> List[int]:
...
+    n = len(lines)
+    while i < n:
+        # Process each line to find function header.
+        line = lines[i]
+        match = re.match(r"(\s*)def\s+\w+", line)
+        if match:
+            base_indent = len(match.group(1))
+            start = i
+            i += 1
+            while i < n:
```
gpsaggese, line 116: Need comment for this block

helpers/hstring.py
```diff
@@ -94,6 +94,48 @@ def get_docstring_line_indices(lines: List[str]) -> List[int]:
...
+                        if lines[i - 1] == "":
+                            i -= 1
+                        else:
+                            break
+                    break
+                i += 1
+            function_line_indices.extend(range(start, i))
+        else:
+            i += 1
+    return function_line_indices
```
gpsaggese, line 136: The code is too complicated, with all the indices back and forth... the logic is not clear.

IMO you can do everything in one pass, scanning one line at the time

Why not just using a state like "look_for_function", "look_for_body" (to skip the comments) and then you keep searching until you find the next function.

linters/amp_remove_empty_lines.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python
```
gpsaggese, line 1: -> amp_remove_empty_lines_in_function.py

linters/amp_remove_empty_lines.py
```diff
@@ -0,0 +1,95 @@
...
+    Process file to find and update functions.
+
+    :param file_content: file to process
+    :return: formatted file without empty lines in functions
+    """
+    lines = file_content.splitlines()
+    if file_content.endswith("\n"):
+        lines.append("")
+    docstring_indices = set(hstring.get_docstring_line_indices(lines))
+    function_indices = set(hstring.get_function_indices(lines))
```
gpsaggese, line 32: let's move get_function_indices here

linters/amp_remove_empty_lines.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def update_function_blocks(file_content: str) -> List[str]:
+    """
+    Process file to find and update functions.
+
+    :param file_content: file to process
+    :return: formatted file without empty lines in functions
+    """
+    lines = file_content.splitlines()
```
gpsaggese, line 28: One comment per code chunk.

linters/amp_remove_empty_lines.py
```diff
@@ -0,0 +1,95 @@
...
+
+import helpers.hdbg as hdbg
+import helpers.hio as hio
+import helpers.hparser as hparser
+import helpers.hstring as hstring
+import linters.action as liaction
+import linters.utils as liutils
+
+
+def update_function_blocks(file_content: str) -> List[str]:
```
gpsaggese, line 21: file_content -> txt

gpsaggese, line 21: update how?

linters/amp_remove_empty_lines.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    Process file to find and update functions.
+
+    :param file_content: file to process
+    :return: formatted file without empty lines in functions
+    """
+    lines = file_content.splitlines()
+    if file_content.endswith("\n"):
```
gpsaggese, line 29: Why do we need an empty line at the end?

linters/amp_remove_empty_lines.py
```diff
@@ -0,0 +1,95 @@
...
+        lines.append("")
+    docstring_indices = set(hstring.get_docstring_line_indices(lines))
+    function_indices = set(hstring.get_function_indices(lines))
+    cleaned_file = []
+    for i, line in enumerate(lines):
+        stripped = line.strip()
+        # Skip empty lines, except in docstrings.
+        if (
```
gpsaggese, line 37: So complicated. Find the docstrings, find the function bodies, ... why can't it be done in a single pass? A normal function has a regular structure

```
def foobar(...)
    """
    """
    blah blah

something else
```

linters/test/test_amp_remove_empty_lines.py
```diff
@@ -0,0 +1,53 @@
...
+        Test cleaning empty lines in methods and functions without any empty
+        lines.
+        """
+        self._run_test()
+
+    def _run_test(self) -> None:
+        """
+        Helper to run test cases.
+        """
+        test_input_dir = self.get_input_dir()
```
gpsaggese, line 44:
```
# Prepare inputs.
```

linters/test/test_amp_remove_empty_lines.py
```diff
@@ -0,0 +1,53 @@
...
+        text_file_path = os.path.join(test_input_dir, "test.txt")
+        text = hio.from_file(text_file_path)
+        # Run.
+        actual = lareemli.update_function_blocks(text)
+        # Check.
+        test_output_dir = self.get_output_dir()
+        output_file_path = os.path.join(test_output_dir, "test.txt")
+        expected = hio.from_file(output_file_path)
+        self.assertEqual(expected, "\n".join(actual))
```
gpsaggese, line 53: -> assert_equal is better

linters/amp_remove_empty_lines.py
allenmatt10: I have renamed it to `_remove_empty_lines()` for more clarity.

helpers/hstring.py
allenmatt10: A single pass was removing empty line(s) succeeding the function, as it couldn't reliably determine where the function ended (since empty lines have no indentation). I will simplify the current implementation and add more comments for clarity.

allenmatt10:
I have updated the code as per the feedback given:

- Moved the `_find_function_indices()` from `helpers.hstring` back to `linters/amp_remove_empty_lines_in_function.py`.
- Added `_LOG.debug` statements for debugging.
- Added more comments and examples for better clarity.

linters/amp_remove_empty_lines_in_function.py
```diff
@@ -0,0 +1,160 @@
...
+import linters.utils as liutils
+
+_LOG = logging.getLogger(__name__)
+
+
+# TODO(allenmatt10): Add tests.
+def _find_function_indices(text: List[str]) -> List[int]:
+    """
+    Get indices of lines of code that are inside functions and methods.
+    For example:
```
sonniki, line 28: As per our conventions, the first line of the docstring should be a single line. So, there should be an empty line in between here.

linters/amp_remove_empty_lines_in_function.py
```diff
@@ -0,0 +1,160 @@
...
+            i += 1
+            continue
+        base_indent = len(match.group(1))
+        start = i
+        _LOG.debug(
+            "Function header found at line ",
+            i,
+            " with base indentation ",
+            base_indent,
+        )
```
sonniki, line 64: Pls grep for `_LOG.debug` in the codebase to see how we construct log messages with inserted variables

sonniki (changes requested):

sonniki:
> I'm assuming it could be related to a recent push (https://github.com/causify-ai/helpers/pull/544) which updated /dev_scripts_helpers/documentation/lint_notes.py
> @sonniki Could this be related, in your opinion?

Indeed, it's a build break in master, unrelated to this PR. Should be fixed soon.

linters/amp_remove_empty_lines_in_function.py
```diff
@@ -0,0 +1,160 @@
...
+import helpers.hio as hio
+import helpers.hparser as hparser
+import helpers.hstring as hstring
+import linters.action as liaction
+import linters.utils as liutils
+
+_LOG = logging.getLogger(__name__)
+
+
+# TODO(allenmatt10): Add tests.
```
sonniki, line 24: @allenmatt10 feel free to already file an issue and assign to yourself (put me/GP as FYI). Ofc we need to wait until this PR is merged to work on it.

sonniki (approved):
LGTM. @gpsaggese okay with you to merge? Or do you want @allenmatt10 to make any more changes?

linters/amp_remove_empty_lines_in_function.py
allenmatt10: Yes. I will file an issue now and work on it after this PR is merged.

linters/amp_remove_empty_lines_in_function.py
```diff
@@ -0,0 +1,160 @@
...
+def _find_function_indices(text: List[str]) -> List[int]:
+    """
+    Get indices of lines of code that are inside functions and methods.
+
+    For example:
+        Input: [
```
gpsaggese, line 30: Let's write it as

```
0: import os
1:
2: def func1(...
3
```

to help the reader

linters/amp_remove_empty_lines_in_function.py
```diff
@@ -0,0 +1,160 @@
...
+            continue
+        base_indent = len(match.group(1))
+        start = i
+        _LOG.debug(
+            "Function header found at line %d with base indentation %d",
+            i,
+            base_indent,
+        )
+        i += 1
+        while i < n:
```
gpsaggese, line 66: ok @sonniki to decide if you want to merge it and then rewrite it replacing the triple nested loop with a single pass and a state.

allenmatt10: @gpsaggese 

I have replaced the triple nested loop. The main challenge was to retain empty lines succeeding the functions as its impossible to differentiate empty lines inside and outside the function, which is why I went with a triple nested loop initially. Let me know what you think of this implementation.

linters/amp_remove_empty_lines_in_function.py
```diff
@@ -0,0 +1,134 @@
...
+                        # Retain empty lines between two functions.
+                        cleaned_file.append("")
+                        i -= 1
+            else:
+                # Retain empty lines between the previous function and the surrounding code.
+                while lines[i - 1] == "":
+                    cleaned_file.append("")
+                    i -= 1
+                inside_function = False
+                base_indent = 0
```
sonniki, line 81: Wouldn't this 
```
                while lines[i - 1] == "":
                    cleaned_file.append("")
                    i -= 1
                inside_function = False
                base_indent = 0
```
also work for the case where we have one function after the other? Unless I'm miscalculating, it would bring `i` to the empty line in between functions, with `inside_function=False` and `base_indent=0`, and then we'll continue from there onto the new function.

sonniki (commented):
TBH I thought the previous implementation was clearer... But it's @gpsaggese's call on our coding style. I'm not going to fight for it:)

linters/amp_remove_empty_lines_in_function.py
allenmatt10: Not really. Lets take an example:

```
<empty line>
def foo():
    
    print("inside foo()")
<empty line> 
def bar():
    
    print("inside bar()")
```
Now, since `def foo()` is the first function we process, the empty lines before it are already registered as it is non function code, so we ignore them here:

```
    while lines[i - 1] == "":
               if cleaned_file[k - 1] == "":
                   # Check if no functions precede the current one,
                   # then the empty lines are already retained.
                   i -= 1
                   k -= 1
```
When we process the empty line between `def foo()` and `def bar()`, the state `inside_function` still remains `True`, and at the next line when it encounters `def bar()`, the state is set to `True` again. So the `if match:` passes, and the empty line is retained here:

```
    else:
             # Retain empty lines between two functions.
              cleaned_file.append("")
              i -= 1
```

If there is any non-function code after the above empty line, then `if match:` fails, and we proceed to retain the empty lines here:
```
                while lines[i - 1] == "":
                    cleaned_file.append("")
                    i -= 1
                inside_function = False
                base_indent = 0
```

I have tested several cases locally, as I wanted to protect all non-function empty lines (which are added/handled by other linter actions).

sonniki:
@gpsaggese okay with you to merge?

sonniki:
> @gpsaggese okay with you to merge?

"Silence means yes" :)
Merging as checkpoint, follow-ups can be filed for refactoring if needed

Issue #494, opened 2025-04-04
Bug in helpers.hmarkdown.is_markdown_line_separator()


sandeepthalapanane:
While implementing test cases ( #462 ) for the helpers.hmarkdown.format_headers() function, I found a bug. The regex pattern `#*\s*-----+` in the `is_markdown_line_separator()` function fails to match lines containing fewer than five dashes. However, we can use a line separator made of triple dashes `---`.

```python 
def is_markdown_line_separator(line: str) -> bool:
    res = (
        re.match(r"#*\s*#####+", line)
        or re.match(r"#*\s*/////+", line)
        or re.match(r"#*\s*-----+", line)
        or re.match(r"#*\s*=====+", line)
    )
    res = bool(res)
    return res
```

@sonniki @samarth9008

sonniki:
I think maybe it was deemed to be too prone to FPs when it's fewer than 5 dashes. @gpsaggese I believe you wrote the code originally?  Is that correct?

gpsaggese:
TBH, the idea is to catch up with situations when people just type a bit and don't align the bar.

I'm ok with anything. The current thing hurts my eyes. I'd do something like

```
t = 5
regex = rf"#*\s*(?:#{t,}|/{t,}|-{t,}|={t,})"
res = re.match(regex, line)
```

or after a ChatGPT treatment

```
t = 5
regex = rf"""
    # Optional leading comment characters
    \#*            

    # Optional whitespace
    \s*            

    # Match at least {t} of any one of the following characters:
    (?:             
        \#{{{t},}}   # At least {t} hash symbols
        | /{{{t},}}  # At least {t} slashes
        | -{{{t},}}  # At least {t} dashes
        | ={{{t},}}  # At least {t} equals signs
    )
"""
res = re.match(regex, line, re.VERBOSE)
```

sonniki:
SG. @sandeepthalapanane feel free to file a PR following GP's suggestions

sandeepthalapanane:
> or after a ChatGPT treatment
> 
> ```
> t = 5
> regex = rf"""
>     # Optional leading comment characters
>     \#*            
> 
>     # Optional whitespace
>     \s*            
> 
>     # Match at least {t} of any one of the following characters:
>     (?:             
>         \#{{{t},}}   # At least {t} hash symbols
>         | /{{{t},}}  # At least {t} slashes
>         | -{{{t},}}  # At least {t} dashes
[...]

PR #534, opened 2025-04-10
Modify helpers.hmarkdown.is_markdown_line_seperator()

sandeepthalapanane:
Following the discussion in #494, I have modified the `helpers.hmarkdown.is_markdown_line_separator()` function and the test cases.


helpers/hmarkdown.py
```diff
@@ -26,20 +26,22 @@
...
 
 
-def is_markdown_line_separator(line: str) -> bool:
+def is_markdown_line_separator(line: str, min_repeats: int = 3) -> bool:
     """
     Check if the given line is a Markdown separator.
 
     :param line: the current line of text being processed
+    :param min_repeats: the minimum number of times the characters are
+        repeated  (`#`, `/`, `-`, `=`) to be considered a separator
```
sonniki, line 35: I would move the presentation of characters above, into the main docstring.

helpers/hmarkdown.py
```diff
@@ -26,20 +26,22 @@
...
+        - E.g., min_repeats = 2 classifies `##`, `###`, `//` as line seperator, but not `#`, `/`
```
sonniki, line 36: Let's do backticks for var names.
"E.g., if `min_repeats` = 2, then `##`, `###`, `//` are considered to be line separators, but `#`, `/` are not"

helpers/hmarkdown.py
```diff
@@ -26,20 +26,22 @@
...
-        re.match(r"#*\s*######+", line)
-        or re.match(r"#*\s*/////+", line)
-        or re.match(r"#*\s*-----+", line)
-        or re.match(r"#*\s*=====+", line)
-    )
-    res = bool(res)
+    line_pattern = rf"""
+    \#*\s*  # Allow optional leading `#` and whitespace.
+    ([#/=\-])\1{{{min_repeats - 1},}}  # Capture a character, then repeat it (`min_repeats` - 1) times.
+    \s*$ # Match only whitespace characters until the end of the line.
```
sonniki, line 42: Generally we don't allow inline comments. Not sure if the regex would break if we move the comments into separate lines (above the line each comment is describing).

helpers/hmarkdown.py
```diff
@@ -26,20 +26,22 @@
...
     :return: true if the line is a separator
     """
-    res = (
-        re.match(r"#*\s*######+", line)
-        or re.match(r"#*\s*/////+", line)
-        or re.match(r"#*\s*-----+", line)
-        or re.match(r"#*\s*=====+", line)
-    )
-    res = bool(res)
+    line_pattern = rf"""
```
sonniki, line 39: -> `separator_pattern`

helpers/test/test_hmarkdown.py
```diff
@@ -172,7 +172,7 @@ def test_valid_separator1(self) -> None:
 
     def test_valid_separator2(self) -> None:
```
sonniki, line 173: Let's also add what we've discussed in the issue to test cases. Stuff like:
```
=================================== FAILURES ===================================
-----BEGIN OPENSSH PRIVATE KEY-----
```

sonniki (changes requested):
Some style things

helpers/hmarkdown.py
sandeepthalapanane: I tried to move above the line, and it didn't break the regex pattern.

sonniki (changes requested):
Small thing, then we can merge IMO

sonniki:
BTW @sandeepthalapanane please don't forget:
- The title of the PR should match the name of the branch (see other PRs in the repo for reference)
- Commit messages should start with the tag of the issue, e.g. `HelpersTask494: Change XYZ"

sonniki (approved):

Issue #445, opened 2025-03-29
Fix bug in removed_blocks_storage linter


gpsaggese:
When I lint several files I'm hitting this problem

i lint --files="helpers/lib_tasks_find.py"

```
Traceback (most recent call last):
  File "/app/helpers_root/./linters/base.py", line 519, in <module>
    _main(args_)
  File "/app/helpers_root/./linters/base.py", line 497, in _main
    lints = _run_linter(file_paths, action_names, action_classes, args)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/helpers_root/./linters/base.py", line 393, in _run_linter
    cur_file_lints = _lint(
                     ^^^^^^
  File "/app/helpers_root/./linters/base.py", line 354, in _lint
    cur_action_lints = action_class.execute(file_path, pedantic)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/helpers_root/linters/action.py", line 62, in execute
    output = self._execute(file_name, pedantic)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/helpers_root/linters/action.py", line 151, in _execute
    output.extend(action.execute(file_name, pedantic))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/helpers_root/linters/action.py", line 62, in execute
    output = self._execute(file_name, pedantic)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/helpers_root/linters/amp_doc_formatter.py", line 202, in _execute
    self._restore_removed_code_blocks(file_name, _removed_code)
  File "/app/helpers_root/linters/amp_doc_formatter.py", line 171, in _restore_removed_code_blocks
    for skipped_line in removed_blocks_storage[skipped_id]:
                        ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
KeyError: '4'
```

I've seen this problem in several files now.

Can we make this part more robust, add tests?

sonniki:
[...]

PR #535, opened 2025-04-10
HelpersTask445_Fix_bug_in_removed_blocks_storage_linter

indrayudd:
Addresses #445


linters/amp_doc_formatter.py
```diff
@@ -49,6 +49,41 @@ def check_if_possible(self) -> bool:
...
+        ]
+        # Count triple backticks from non-comment lines.
+        backticks = []
+        for line in non_comment_lines:
+            backticks.extend(re.findall(r"```", line))
+        if len(backticks) % 2 != 0:
+            # Unbalanced backticks.
+            # Locate first non-comment line with triple backticks.
+            for i, line in enumerate(lines, start=1):
+                if not line.lstrip().startswith("#") and "```" in line:
```
sonniki, line 76: It would be best to use functions that we already have, like `hstring.get_docstring_line_indices()` to get the docstring lines, and `hstring.get_code_block_line_indices()` to get code blocks within those lines. You can modify these functions to generalize them to fit the purpose (although if you do so, make sure to not break the behavior in places where they are already used).

linters/amp_doc_formatter.py
```diff
@@ -49,6 +49,41 @@ def check_if_possible(self) -> bool:
...
+                    print(True, i)
+                    return True, i
+        # Check each non-comment line for misplaced triple backticks.
+        for i, line in enumerate(lines, start=1):
+            if line.lstrip().startswith("#"):
+                continue
+            if "```" in line and not line.lstrip().startswith("```"):
+                return True, i
```
sonniki, line 84: IMO we don't need to separate these two cases. We just detect when triple backticks were opened but not closed, and whatever the reason for this, we don't care, we just issue a general warning through Linter, smth like `Docstring contains backticks that are unbalanced or misplaced; make sure both opening and closing backticks are the leftmost element of their line`.

linters/amp_doc_formatter.py
```diff
@@ -49,6 +49,41 @@ def check_if_possible(self) -> bool:
...
+        return False, 0
```
sonniki, line 85: Such returns are not very clear. It's better to assign these values to some variables, like `_detected_unbalanced_backticks` and `line_num`, and then return the variables

linters/amp_doc_formatter.py
```diff
@@ -188,17 +223,32 @@ def _execute(self, file_name: str, pedantic: int) -> List[str]:
         if self.skip_if_not_py(file_name):
             # Apply only to Python files.
             return []
+        # Check for misplaced backticks
```
sonniki, line 226: All comments must end with `.`

linters/amp_doc_formatter.py
```diff
@@ -188,17 +223,32 @@ def _execute(self, file_name: str, pedantic: int) -> List[str]:
...
+        misplaced_warning = ""
+        has_misplaced, line_no = self._has_misplaced_triple_backticks(file_name)
         # Clear and store ignored docstrings and code.
         _ignored_docstrings = self._remove_ignored_docstrings(file_name)
-        _removed_code = self._remove_code_blocks(file_name)
+        if has_misplaced:
+            # Do not remove any codeblocks as misplaced/unbalanced backticks detected.
+            misplaced_warning = (
+                f"{file_name}:{line_no}: "
```
sonniki, line 234: BTW it's okay if the line number doesn't point to the exact line with the open backticks that are not closed afterwards. It's enough to point at the first line of the docstring where this is happening. Just see what is possible.

linters/test/test_amp_doc_formatter.py
```diff
@@ -132,6 +132,50 @@ def test6(self) -> None:
...
+            file_name=temp_file, pedantic=0
+        )[0]
+        # Check.
+        expected = (
+            f"{temp_file}:2: Found misplaced or unbalanced triple backticks"
+        )
+        self.assertEqual(actual, expected)
+
+        # Balanced backticks.
+        # Prepare inputs.
```
sonniki, line 162: This very clearly should be two tests instead of one

linters/test/test_amp_doc_formatter.py
```diff
@@ -132,6 +132,50 @@ def test6(self) -> None:
...
+        text = '''content = r"""
+```python
+def no_closing_backticks():
+    print("No below delimiters are present")
```
sonniki, line 166: This one does have closing backticks

linters/test/test_amp_doc_formatter.py
```diff
@@ -132,6 +132,50 @@ def test6(self) -> None:
...
+        temp_file = os.path.join(scratch_dir, "temp_file.py")
+        hio.to_file(temp_file, text)
+        # Run.
+        actual = lamdofor._DocFormatter().execute(
+            file_name=temp_file, pedantic=0
+        )[0]
+        # Check.
+        expected = (
+            f"{temp_file}:2: Found misplaced or unbalanced triple backticks"
+        )
```
sonniki, line 158: We should also check the file itself after the docformatter, to check that it's no longer being incorrectly modified

linters/test/test_amp_doc_formatter.py
```diff
@@ -132,6 +132,50 @@ def test6(self) -> None:
         actual = self._docformatter(text)
         self.assertEqual(expected, actual)
 
+    def test7(self) -> None:
```
sonniki, line 135: There should also be a test representing the situation from `helpers/lib_tasks_find.py`

sonniki (changes requested):

linters/amp_doc_formatter.py
indrayudd: Would an output like `linters/test_helper.py:12-17: Found unbalanced triple backticks;  make sure both opening and closing backticks are the leftmost element of their line [doc_formatter]` look good? It basically flags all the blocks of docstrings with the backtick discrepancy.

sonniki: Hmm, I would still keep a single line reference, e.g. `12` instead of `12-17`. Is it possible? See also the note in https://github.com/causify-ai/helpers/pull/535#discussion_r2039243399

indrayudd: Yeah, as mentioned, the starting index of every docstring with a discrepancy will be possible to implement, working on it.

linters/amp_doc_formatter.py
```diff
@@ -49,6 +49,41 @@ def check_if_possible(self) -> bool:
...
+        """
+        Check if the file contains triple backticks that are either unbalanced
+        or misplaced.
+
+        :param file_name: file to process
+        :return: true if misplaced/unbalanced backticks are detected,
+            line where first issue occurs
+        """
+        contents = hio.from_file(file_name)
+        lines = contents.splitlines()
```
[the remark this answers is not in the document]
indrayudd: Factoring out `contents = hio.from_file(file_name)` into `_execute` partially works; when I pass `contents = hio.from_file(file_name)` into `_restore_ignored_docstrings()` and `_remove_code_blocks()` as parameters instead of calling the `hio.from_file()` function inside them, most of the tests fail. I could successfully pass `contents` into the remaining functions, though, reducing redundancy to some extent. I'm not sure why that is, I can debug this further once my primary deliverables are submitted.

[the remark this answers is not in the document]
indrayudd: Update: @sonniki, turns out we can't factor this out in the `_execute` definition because almost every function in the flow makes changes to the file; so we need to read it multiple times to keep the content the latest. This is why the python related slow tests failed in the recent PR.

linters/amp_doc_formatter.py
```diff
@@ -49,12 +49,54 @@ def check_if_possible(self) -> bool:
         check: bool = hsystem.check_exec(self._executable)
         return check
 
+    @staticmethod
+    def _has_unbalanced_triple_backticks(file_name: str) -> Tuple[bool, List]:
+        """
+        Check if the file contains triple backticks that are unbalanced.
```
gpsaggese, line 55: Let's clarify what "unbalanced" means. Things are clear now that we have context, but in 6 months we will have forgotten

Let's explain why we need to find them. Many times what's unclear is how a function can be used, rather than what it does

linters/amp_doc_formatter.py
```diff
@@ -49,12 +49,54 @@ def check_if_possible(self) -> bool:
...
+
+        :param contents: contents of the file to process
+        :return:
+            - whether there are unbalanced backticks
+            - all the starting indices of docstrings where the flagged backticks exist
```
gpsaggese, line 60: flagged?

linters/amp_doc_formatter.py
```diff
@@ -49,12 +49,54 @@ def check_if_possible(self) -> bool:
         check: bool = hsystem.check_exec(self._executable)
         return check
 
+    @staticmethod
```
gpsaggese, line 52: I am not fully in the loop, but why not having a function that parses the code looking for the blocks enclosed in triple backticks, and if the number is not even we assert.

In other words in any other parts of the codebase we look for the chunks (which is the primary reason) and then we assert if there is a problem.

linters/amp_doc_formatter.py
```diff
@@ -188,17 +232,34 @@ def _execute(self, file_name: str, pedantic: int) -> List[str]:
         if self.skip_if_not_py(file_name):
             # Apply only to Python files.
             return []
+        # Check for unbalanced backticks.
+        unbalanced_backtick_warning = ""
+        has_misplaced, issues = self._has_unbalanced_triple_backticks(file_name)
```
gpsaggese, line 237: misplaced? We call them unbalanced in other places

linters/test/test_amp_doc_formatter.py
```diff
@@ -132,6 +132,100 @@ def test6(self) -> None:
...
+        - `print_deps`: print the result of grepping for imports
+        - `find_deps`: find all the dependencies
+        - `find_lev1_deps`, `find_lev2_deps`: find all the dependencies
+    :param only_module: keep only imports containing a certain module (e.g., `dataflow`)
+    :param ignore_standard_libs: ignore the Python standard libs (e.g., `os`, `...`)
+    :param ignore_helpers: ignore the `helper` lib
+    :param remove_dups: remove the duplicated imports
+    """
+'''
+        scratch_dir = self.get_scratch_space()
```
gpsaggese, line 212: Let's factor out the redundancy of the test

linters/test/test_amp_doc_formatter.py
```diff
@@ -132,6 +132,100 @@ def test6(self) -> None:
...
+
+    ```
+    Missing closing backticks.
+    """
+'''
+        scratch_dir = self.get_scratch_space()
+        temp_file = os.path.join(scratch_dir, "temp_file.py")
+        hio.to_file(temp_file, text)
+        # Run.
+        actual_warnings = lamdofor._DocFormatter().execute(
```
gpsaggese, line 171: The testing is too end-to-end. Can we just test _has_unbalanced_triple_backticks?

linters/test/test_amp_doc_formatter.py
indrayudd: By factoring out the redundancy, do you mean having one test check `_has_unbalanced_triple_backticks` and the other the way it is - an end to end test of the excerpt from `lib_tasks_find.py` that previously broke the linter, as @sonniki suggested?

linters/amp_doc_formatter.py
indrayudd: I think the function is doing what you're suggesting already. It:

- gets every chunk of docstring in the script (the primary trigger for the bug)
- looks for odd number of backticks in every chunk and return the starting indices of the ones that have an odd number of triple backticks

linters/amp_doc_formatter.py
indrayudd: My bad, changed it to unbalanced.

linters/amp_doc_formatter.py
indrayudd: Changed it to unbalanced for clarity.

linters/amp_doc_formatter.py
indrayudd: Added an example and the purpose.

linters/test/test_amp_doc_formatter.py
indrayudd: Added this in `test8()`

linters/amp_doc_formatter.py
```diff
@@ -188,17 +241,34 @@ def _execute(self, file_name: str, pedantic: int) -> List[str]:
...
         _ignored_docstrings = self._remove_ignored_docstrings(file_name)
-        _removed_code = self._remove_code_blocks(file_name)
+        if is_unbalanced:
+            # Do not remove any codeblocks as unbalanced backticks detected.
+            unbalanced_backtick_warning = "\n".join(
+                f"{file_name}:{start}: Found unbalanced triple backticks; "
+                f"make sure both opening and closing backticks "
+                f"are the leftmost element of their line"
+                for start in issues
+            )
```
sonniki, line 256: This can be moved to the end of the function (and then we can get rid of `unbalanced_backtick_warning` var)

linters/amp_doc_formatter.py
```diff
@@ -188,17 +241,34 @@ def _execute(self, file_name: str, pedantic: int) -> List[str]:
...
         # Clear and store ignored docstrings and code.
         _ignored_docstrings = self._remove_ignored_docstrings(file_name)
-        _removed_code = self._remove_code_blocks(file_name)
+        if is_unbalanced:
+            # Do not remove any codeblocks as unbalanced backticks detected.
+            unbalanced_backtick_warning = "\n".join(
+                f"{file_name}:{start}: Found unbalanced triple backticks; "
+                f"make sure both opening and closing backticks "
+                f"are the leftmost element of their line"
+                for start in issues
```
sonniki, line 255: `start` -> `start_idx`

linters/amp_doc_formatter.py
```diff
@@ -188,17 +241,34 @@ def _execute(self, file_name: str, pedantic: int) -> List[str]:
...
         cmd = f"{self._executable} {opts} {file_name}"
         output: List[str] = []
         _, output = liutils.tee(cmd, self._executable, abort_on_error=False)
         # Restore ignored docstrings and code.
         self._restore_ignored_docstrings(file_name, _ignored_docstrings)
-        self._restore_removed_code_blocks(file_name, _removed_code)
+        if _removed_code:
+            self._restore_removed_code_blocks(file_name, _removed_code)
+        if unbalanced_backtick_warning:
+            # Prepend generated warnings.
```
sonniki, line 270: IMO no need to prepend necessarily. Can just append (or extend) and have these warnings after the ones coming from `docformatter`

linters/test/test_amp_doc_formatter.py
```diff
@@ -132,6 +132,100 @@ def test6(self) -> None:
         actual = self._docformatter(text)
         self.assertEqual(expected, actual)
 
+    def test7(self) -> None:
+        """
+        Test that unbalanced backticks are correctly flagged.
+        """
+        # Prepare inputs.
+        text = '''def func1():
```
sonniki, line 140: I'd move this into a file in `input` for better readability (see [here](https://github.com/causify-ai/helpers/blob/8fa9ff8905d76310e2641126a9291392ea6be0bc/docs/coding/all.write_unit_tests.how_to_guide.md#our-framework-to-test-using-input--output-data))

sonniki (changes requested):

linters/test/test_amp_doc_formatter.py
sonniki: Agreed, let's make this test only call that specific function

linters/test/test_amp_doc_formatter.py
sonniki: IMO we can still have this test here to verify the bug fix end-to-end (although it's not necessary to have it completely duplicate the docstring from `lib_tasks_find.py`, we can make it more compact while keeping the structure of it). 

However, @indrayudd note that the other tests in this file use `_docformatter()` where all the infrastructure around creating/reading temp files and running `docformatter` is factored out. I believe this is what "factoring out the redundancy" means - but actually you don't have to factor it out, it already is, you just have to use the helper.

sonniki:
@indrayudd don't forget to resolve conversations when you've addressed the comments -- it helps track the progress

linters/test/test_amp_doc_formatter.py
indrayudd: In that case, I'll change `_docformatter()` to also return warnings and change all the tests accordingly.

linters/amp_doc_formatter.py
```diff
@@ -49,6 +49,58 @@ def check_if_possible(self) -> bool:
...
+        # Process each chunk.
+        for chunk in chunks:
+            leftmost_count = 0
+            for idx in chunk:
+                if lines[idx].lstrip().startswith("```"):
+                    # Count this triple backticks at the leftmost position.
+                    leftmost_count += 1
+            if leftmost_count % 2 != 0:
+                # Flag the chunk.
+                # Convert zero-indexed numbers to one-indexed line numbers.
```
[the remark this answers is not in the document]
indrayudd: This is because the line number on editors are one-indexed. Mentioning that in an additional comment.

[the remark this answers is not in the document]
sonniki: I don't think we do it in other Linter steps. Could you please quickly go over them and confirm? We want to keep it consistent with other Linter steps (including ones where we don't have control over the warning construction because they come from a 3rd-party tool). If in the other steps we don't convert from zero-indexed to one-indexed, then let's also keep it zero-indexed here.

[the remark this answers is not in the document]
indrayudd: In my interaction with other Linter warnings, every time I've clicked on a link to the `script.py:line_number`, it's always accurately opened on that line on the editor. If it were zero indexed, wouldn't it open on the line after the flagged line? Doing a quick check regardless.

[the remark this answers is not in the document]
indrayudd: From digging around a little I found out that one of the arguments passed to `pylint` is `--output-format=parseable` which returns line numbers compatible with programming tools, aka, one indexed lines.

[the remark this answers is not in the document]
sonniki: Okay. But what about those scripts that are not wrappers around 3rd-party tool, such as `pylint`, but the ones that we custom-wrote? Just as an example, in `amp_warn_incorrectly_formatted_todo.py` and `amp_fix_md_links.py`, the warnings show the zero-indexed line num. We want to have consistent behavior in that regard -- could you please file an issue to investigate this in all the Linter scripts and unify? You can assign it to yourself and put me and GP as FYI.

linters/amp_doc_formatter.py
```diff
@@ -49,6 +49,72 @@ def check_if_possible(self) -> bool:
         check: bool = hsystem.check_exec(self._executable)
         return check
 
+    @staticmethod
+    def _get_docstring_lines(lines: List[str]) -> List[List[int]]:
+        """
+        Return the line indices that are within docstrings.
```
sonniki, line 55: From this it's not clear what is the difference between this function and `hstring.get_docstring_line_indices()`. Better to say smth like "Get line indices of individual docstrings", or something similar, to emphasize that individual docstrings are separated in the output.

linters/amp_doc_formatter.py
```diff
@@ -49,6 +49,72 @@ def check_if_possible(self) -> bool:
...
+
+        :param lines: lines from the file to process
+        :return: lines within docstrings
```
sonniki, line 58: Same here

linters/amp_doc_formatter.py
```diff
@@ -49,6 +49,72 @@ def check_if_possible(self) -> bool:
...
+        :param file_name: file to process
+        :return:
+            - whether there are unbalanced backticks
+            - all the starting indices of docstrings where the
+              unbalanced backticks exist
+        """
+        contents = hio.from_file(file_name)
+        lines = contents.splitlines()
+        # Get lines that are within docstrings.
+        docstrings = _DocFormatter._get_docstring_lines(lines)
```
sonniki, line 100: This is unexpected:) Is this done so that this method can be called without initializing `_DocFormatter` in the test? Is it a problem if we initialize it? IMO it would be better to initialize the class there than to make this method static like this. Another solution (which I would prefer) is to move `_get_docstring_lines()` to `hstring`. It's really more general purpose than this class and thematically it should belong in the same place as `hstring.get_docstring_line_indices()`

linters/amp_doc_formatter.py
```diff
@@ -49,6 +49,72 @@ def check_if_possible(self) -> bool:
...
+        idxs_docstrings_with_unbalanced_backticks = []
+        # Process each docstring.
+        for docstring in docstrings:
+            leftmost_triple_backticks_count = 0
+            for idx in docstring:
+                if lines[idx].lstrip().startswith("```"):
+                    # Count this triple backticks at the leftmost position.
+                    leftmost_triple_backticks_count += 1
+            if leftmost_triple_backticks_count % 2 != 0:
+                # Odd number of leftomost triple backticks in this docstring.
```
sonniki, line 110: typo, -> "leftmost"

linters/test/test_amp_doc_formatter.py
```diff
@@ -129,19 +130,81 @@ def test6(self) -> None:
         text_file_path = os.path.join(test6_input_dir, "test.txt")
         text = hio.from_file(text_file_path)
         expected = text
-        actual = self._docformatter(text)
+        actual, _, _ = self._docformatter(text)
         self.assertEqual(expected, actual)
 
-    def _docformatter(self, text: str) -> str:
+    def test_get_docstring_lines(self) -> None:
```
sonniki, line 136: Our convention is to have a separate test class for testing each separate function/method (and then testing functions `test1`, `test2`, etc, within the test class).
The `_docformatter()` method can be moved out of the class and turned into a function, which will then be used by several classes.

linters/test/test_amp_doc_formatter.py
```diff
@@ -129,19 +130,81 @@ def test6(self) -> None:
...
+        expected = [
+            [1, 2, 3, 4, 5, 6],
+            [11, 12, 13, 14, 15],
+            [20, 21, 22, 23, 24],
+        ]
+        self.assertEqual(actual, expected)
+
+    def test_find_unbalanced_triple_backticks(self) -> None:
+        """
+        Test that the starting idx of docstrings with unbalanced backticks are
```
sonniki, line 159: "idx" -> "indices"

linters/test/outcomes/Test_docformatter.test9/input/test.txt
```diff
@@ -0,0 +1,12 @@
+
+    """
```
sonniki, line 2: Not sure we need to move it to the `input` dir, IMO we can make it more compact while keeping the problematic structure (i.e. instead of the original text write some short placeholders but keep the structure of where the backticks are) and move it back to the test. WDYT?

linters/test/test_amp_doc_formatter.py
```diff
@@ -129,19 +130,81 @@ def test6(self) -> None:
...
+
+    def _docformatter(self, text: str) -> Tuple[str, List[str], str]:
         """
         Run the docformatter on the temp file in scratch space.
 
         :param text: content to be formatted
-        :return: modified content after formatting
+        :return:
+            - modified content after formatting
+            - warnings array
```
sonniki, line 200: -> "warnings" (no need to specify that it's an array)

sonniki (changes requested):
Mostly small things. Almost there!

linters/test/outcomes/Test_docformatter.test9/input/test.txt
indrayudd: Yeah that sounds good and doable!

linters/test/test_amp_doc_formatter.py
indrayudd: I'm passing `scratch_dir` as a parameter in order to make `_docformatter()` a function outside of the class, does that work?

helpers/test/outcomes/TestGetDocstrings.test1/input/test.txt
```diff
@@ -0,0 +1,26 @@
+def func1():
+    """
+    First function.
+
+    ```
+    Valid backticks.
```
sonniki, line 6: No need to have valid/invalid backticks in docstrings for this test

linters/test/test_amp_doc_formatter.py
```diff
@@ -129,19 +148,48 @@ def test6(self) -> None:
...
+        file_path = os.path.join(test_get_docstring_lines_input_dir, "test.txt")
+        # Run.
+        actual = lamdofor._DocFormatter._find_unbalanced_triple_backticks(
+            file_path
+        )
+        # Check.
+        expected = [12, 21]
+        self.assertEqual(actual, expected)
+
+    def test9(self) -> None:
```
sonniki, line 170: -> `test7`

sonniki (changes requested):

linters/test/test_amp_doc_formatter.py
sonniki: I think actually that it doesn't need to be moved out of the class after all -- since we only need a separate class for testing `_find_unbalanced_triple_backticks()` and it doesn't require calling the whole `docformatter` script. So it can be moved back in, and the change with the `scratch_dir` reverted

linters/test/test_amp_doc_formatter.py
```diff
@@ -1,10 +1,29 @@
 import os
+from typing import List, Tuple
 
 import helpers.hio as hio
 import helpers.hunit_test as hunitest
 import linters.amp_doc_formatter as lamdofor
 
 
+def _docformatter(text: str, scratch_dir: str) -> Tuple[str, List[str], str]:
```
sonniki, line 9: You might have missed the comment here: https://github.com/causify-ai/helpers/pull/535#discussion_r2048465394

sonniki (changes requested):
Last thing, and then we'll merge

linters/test/test_amp_doc_formatter.py
indrayudd: Just noticed and fixed

sonniki (approved):
All good
(For the future, no need to write the whole branch name in commit messages, just the tag is enough, e.g. `HelpersTask445: Rearrange test files`)

PR #536, opened 2025-04-10
CmampTask11853_Build_failure___cmamp___Update_helpers_submodule___14370314704

heanhsok:
Task [#11853](https://github.com/causify-ai/cmamp/issues/11853)


samarth9008 (commented):
LG but how come this change is needed now

samarth9008:
Yes I get that but how come these tests use to pass till now and suddenly we need to handle them differently. Do we know if we make any changes to the code?

heanhsok:
> Yes I get that but how come these tests use to pass till now and suddenly we need to handle them differently. Do we know if we make any changes to the code?

Yes there was this change in this PR [helpers/hobject.py](https://github.com/causify-ai/helpers/pull/504/files#diff-bf600dd717987c04fb154dd73e2faa250eee2f84b301ca3645d45dc77a9f158a)

Previously, the `_LOG.debug` statement was not executed (only the `_LOG.info` was executed in the CI). Thus we didn't catch this error before.

heanhsok:
> > Yes I get that but how come these tests use to pass till now and suddenly we need to handle them differently. Do we know if we make any changes to the code?
> 
> Yes there was this change in this PR [helpers/hobject.py](https://github.com/causify-ai/helpers/pull/504/files#diff-bf600dd717987c04fb154dd73e2faa250eee2f84b301ca3645d45dc77a9f158a)
> 
> Previously, the `_LOG.debug` statement was not executed (only the `_LOG.info` was executed in the CI). Thus we didn't catch this error before.

Also should we revert this back to just printing out the _LOG.info? The reason is that if we don't limit this to just _LOG.info, it's printing out a lot of stuff which make things very slow when we open the log in the GH Actions. 

@samarth9008

samarth9008:
> Also should we revert this back to just _LOG.info (meaning only messages of level INFO or higher such as WARNING, ERROR, CRITICAL)? The reason is that if we don't limit this to just _LOG.info and above, it's printing out a lot of stuff from _LOG.DEBUG which make things very slow when we open the log in the GH Actions.

Yes it should be just _LOG.info. debug would print lot of noise

samarth9008:
So does it set by default to debug mode since commented out `#_LOG.setLevel(logging.INFO)`

gpsaggese:
> So does it set by default to debug mode since commented out `#_LOG.setLevel(logging.INFO)`

Yes let's reinsert those commented calls to reduce the chattiness of some modules.

I'll think of a better solution. There are debug statements and super chatty debug ones.
I've introduced a level even below DEBUG (TRACE?) but we never used it consistently.
I'll file a bug.

heanhsok:
Merging this for now to unblock others.

I'm skipping one of the test and adding it as a todo as it'll take some time to debug/resolve the problem related to dockerized executable issue and we're not sure if we want to run it in CI as other dockerized executable tests are skipped as well. See https://github.com/causify-ai/helpers/issues/553

Issue #347, opened 2025-03-10
Assert if Linter is not run from repo root


sonniki:
Currently if Linter is run not from the repo root, it will fail because it will compose an incorrect path to the env file and won't be able to find it. E.g. running from `docs/`:
```
Couldn't find env file: /data/sonyan/src/helpers1/docs/devops/env/default.env
```
We should put an assertion in the code to make it clearer for the user that the error is because Linter is run from the wrong directory and they should just run from the repo root to fix it.

gpsaggese:
Makes sense.

The only thing is that sometimes I run from the root of `cmamp` and sometimes from the root of `helpers` and they both work (although probably pick a different file but it doesn't make a difference). 

In any case I'm ok to force running from the top of the outermost repo.

FYI @heanhsok

sonniki:
> The only thing is that sometimes I run from the root of cmamp and sometimes from the root of helpers and they both work (although probably pick a different file but it doesn't make a difference).

Yes, correct, Linter will run fine if it is run from the root of the repo one is developing in, be it `helpers`, `cmamp`, or any other repo.

sonniki:
@Peeyush4 helpful functions regarding checking for repo root should be found in [`helpers/hgit.py`](https://github.com/causify-ai/helpers/blob/master/helpers/hgit.py)

gpsaggese:
Maybe the fix is 

create a function like

```
def is_cwd_git_repo():
   return os.path.exists(".git")
```

and then a dassert on that

The invariant is that in any repo (super and sub) there is a `git` file or dir.

sonniki:
Done

PR #538, opened 2025-04-11
HelpersTask347 Assert if Linter is not run from repo root

Peeyush4:
Task #347

### DIFF
--- a/helpers/lib_tasks_lint.py
+++ b/helpers/lib_tasks_lint.py
@@ -221,6 +221,11 @@ def lint(  # type: ignore
     :param only_format: run only the modifying actions of Linter (e.g., black)
     :param only_check: run only the non-modifying actions of Linter (e.g., pylint)
     """
+    # Check if the user is in a git repo.
+    hdbg.dassert(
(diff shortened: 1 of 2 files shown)


helpers/lib_tasks_lint.py
```diff
@@ -221,6 +221,11 @@ def lint(  # type: ignore
     :param only_format: run only the modifying actions of Linter (e.g., black)
     :param only_check: run only the non-modifying actions of Linter (e.g., pylint)
     """
+    # Check if the user is in a git repo root dir.
```
sonniki, line 224: I would replace "git repo root dir" -> "repo root", for brevity.

helpers/lib_tasks_lint.py
```diff
@@ -221,6 +221,11 @@ def lint(  # type: ignore
...
+    hdbg.dassert(
+        hgit.is_cwd_git_repo(),
+        msg="Linter should run from git repo root dir",
```
sonniki, line 227: Same here

sonniki (changes requested):
LG. Minor change

PR #546, opened 2025-04-11
CmampTask11872_Make_tests_work_on_itsavvy_runnable_dir

heanhsok:
Task CmampTask11872

- Prototype discussed in https://github.com/causify-ai/cmamp/issues/11793#issuecomment-2795577325

### DIFF
--- a/helpers/lib_tasks_docker_release.py
+++ b/helpers/lib_tasks_docker_release.py
@@ -21,6 +21,7 @@ import helpers.lib_tasks_docker as hlitadoc
 import helpers.lib_tasks_pytest as hlitapyt
 import helpers.lib_tasks_utils as hlitauti
 import helpers.repo_config_utils as hrecouti
+import helpers.hio as hio
 
 _DEFAULT_TARGET_REGISTRY = "aws_ecr.ck"
 _LOG = logging.getLogger(__name__)
@@ -613,6 +614,19 @@ def docker_build_prod_image(  # type: ignore
     opts = "--no-cache" if not cache else ""
     # Use dev version for building prod image.
     dev_version = hlitadoc.to_dev_version(prod_version)
+    # Copy source code.
+    current_dir = _to_abs_path(".")
+    helpers_root = hgit.find_helpers_root()
+    helpers_dir = _to_abs_path(helpers_root)
+    deployment_dir = "tmp.deployment"
+    hio.create_dir(deployment_dir, incremental=False)
+    # Copy the source code and deference all the symbolic links.
+    cmd = f"rsync -rL --exclude='{deployment_dir}' {current_dir}/* ./{deployment_dir}"
+    hsystem.system_to_string(cmd)
+    # Copy helpers.
+    cmd = f"cp -rL {helpers_dir}/helpers ./tmp.deployment/"
+    hsystem.system_to_string(cmd)
+    #
     cmd = rf"""
     DOCKER_BUILDKIT={DOCKER_BUILDKIT} \
     time \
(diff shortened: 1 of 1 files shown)


helpers/lib_tasks_docker_release.py
```diff
@@ -613,6 +614,19 @@ def docker_build_prod_image(  # type: ignore
...
+    helpers_dir = _to_abs_path(helpers_root)
+    deployment_dir = "tmp.deployment"
+    hio.create_dir(deployment_dir, incremental=False)
+    # Copy the source code and deference all the symbolic links.
+    cmd = f"rsync -rL --exclude='{deployment_dir}' {current_dir}/* ./{deployment_dir}"
+    hsystem.system_to_string(cmd)
+    # Copy helpers.
+    cmd = f"cp -rL {helpers_dir}/helpers ./tmp.deployment/"
+    hsystem.system_to_string(cmd)
+    #
```
samarth9008, line 629: The solution might work but its very specific to ITSavvy kinda use case. We also use this invoke for lets say building cmamp prod image. In those cases we don't need to this modifications. I also wonder what kind of dir structure it will create inside the container

samarth9008 (commented):
just a thought lmk what you think

helpers/lib_tasks_docker_release.py
```diff
@@ -613,6 +614,19 @@ def docker_build_prod_image(  # type: ignore
     opts = "--no-cache" if not cache else ""
     # Use dev version for building prod image.
     dev_version = hlitadoc.to_dev_version(prod_version)
+    # Copy source code.
+    current_dir = _to_abs_path(".")
+    helpers_root = hgit.find_helpers_root()
+    helpers_dir = _to_abs_path(helpers_root)
+    deployment_dir = "tmp.deployment"
```
gpsaggese, line 621: -> tmp.docker_build_prod_image

helpers/lib_tasks_docker_release.py
```diff
@@ -613,6 +614,19 @@ def docker_build_prod_image(  # type: ignore
...
+    hio.create_dir(deployment_dir, incremental=False)
+    # Copy the source code and deference all the symbolic links.
+    cmd = f"rsync -rL --exclude='{deployment_dir}' {current_dir}/* ./{deployment_dir}"
+    hsystem.system_to_string(cmd)
```
gpsaggese, line 625: why system_to_string a not just system?

helpers/lib_tasks_docker_release.py
```diff
@@ -613,6 +614,19 @@ def docker_build_prod_image(  # type: ignore
...
+    # Copy helpers.
+    cmd = f"cp -rL {helpers_dir}/helpers ./tmp.deployment/"
+    hsystem.system_to_string(cmd)
```
gpsaggese, line 628: Same

helpers/lib_tasks_docker_release.py
```diff
@@ -613,6 +614,19 @@ def docker_build_prod_image(  # type: ignore
...
+    current_dir = _to_abs_path(".")
+    helpers_root = hgit.find_helpers_root()
+    helpers_dir = _to_abs_path(helpers_root)
+    deployment_dir = "tmp.deployment"
+    hio.create_dir(deployment_dir, incremental=False)
+    # Copy the source code and deference all the symbolic links.
+    cmd = f"rsync -rL --exclude='{deployment_dir}' {current_dir}/* ./{deployment_dir}"
+    hsystem.system_to_string(cmd)
+    # Copy helpers.
+    cmd = f"cp -rL {helpers_dir}/helpers ./tmp.deployment/"
```
gpsaggese, line 627: f"... {deployment_dir}"

helpers/lib_tasks_docker_release.py
heanhsok: This change would make the dir structure inside the container looks like
```bash
...
./helpers/...
``` 

Yes this solution is specific for runnable dir. Maybe we can add an if statement and only do this for runnable dir?

samarth9008:
For now I would not invest much time in it until have a proper plan on how to resolve the docker image issues

Issue #464, opened 2025-03-31
Use is_header in hmarkdown


sonniki:
Addressing the [TODO](https://github.com/causify-ai/helpers/blob/bff9cfc764cbad5bfd9f93c0b0e9614eb324fae5/helpers/hmarkdown.py#L414C25-L414C34) in `helpers/hmarkdown.py` of using `is_header()` from the same file instead of using regex directly

sonniki:
Done

PR #552, opened 2025-04-12
HelpersTask464-use-is_header-in-hmarkdown

neomisule:
Link to the issue: https://github.com/causify-ai/helpers/issues/464

Changed helpers/hmarkdown.py code to use is_helper function instead of using regex directly.

FYI @sonniki @samarth9008 @gpsaggese


neomisule:
Not sure why the slow test is failing. I ran the `i run_slow_tests` command to check but it didn't show any fails, all cases passed on my branch as well as the master branch.

helpers/hmarkdown.py
```diff
@@ -408,23 +408,17 @@ def extract_headers_from_markdown(
...
-        # TODO(gp): Use is_header
-        match = header_pattern.match(line)
-        if match:
-            # The number of '#' determines level.
-            level = len(match.group(1))
-            if level <= max_level:
-                title = match.group(2).strip()
-                header_info = HeaderInfo(level, title, line_number)
-                header_list.append(header_info)
+        # Get the level and title from is_header
```
samarth9008, line 417: Dot at the end and I would change the comment bcz mentioning function name again in title is violation of DRY

```
# Get the header level and title.
```

samarth9008 (commented):
Code style review

helpers/hmarkdown.py
```diff
@@ -408,23 +408,17 @@ def extract_headers_from_markdown(
...
+        is_header_bool, level, title = is_header(line)
```
gpsaggese, line 418: is_header_bool -> is_header_

neomisule: Okay, but wont is_header_bool give more context to what we are getting? I thought it will be an easier interpretation for someone seeing that block of code for the first time.

helpers/hmarkdown.py
neomisule: Got it.

helpers/hmarkdown.py
sonniki: We generally avoid using references to the type of a variable in its name, so that if we decide to change implementation later on and change the type of the variable, we don't need to change the name as well. See the [docs](https://github.com/causify-ai/helpers/blob/master/docs/coding/all.coding_style.how_to_guide.md#general-naming-rules): "The name should capture what the data structure represents (its semantics) and not how it is implemented"

Issue #533, opened 2025-04-10
Incorrect header level update after code blocks


Peeyush4:
Linter is automatically changing `###` to `##` for headings that are immediately below the code block, probably confusing a comment in the code with a header of level 1. See https://github.com/causify-ai/helpers/issues/513#issuecomment-2791421226

To reproduce: run Linter on [`docs/work_tools/ck.veracrypt.how_to_guide.md`](https://github.com/causify-ai/helpers/blob/master/docs/work_tools/ck.veracrypt.how_to_guide.md).
Once Linter runs, it converts the header levels of `Create a VeraCrypt volume with non-interactive mode`, `One time set-up` and `Create the tmux env` from `###` to `##`. It's most likely coming from [`linters/amp_check_md_toc_headers.py`](https://github.com/causify-ai/helpers/blob/master/linters/amp_check_md_toc_headers.py).

FYI @sonniki @samarth9008 @gpsaggese

gpsaggese:
Good repro. For me you can even fix it if you don't have anything else to do

Let's follow the process here https://github.com/causify-ai/helpers/issues/280#issuecomment-2794908280

Repro the issue with a unit test, then you can fix it

sonniki:
Done

PR #555, opened 2025-04-14
HelpersTask533 Incorrect header level update after code blocks

Peeyush4:
Task #533


linters/amp_check_md_toc_headers.py
```diff
@@ -26,6 +26,30 @@
 HEADER_REGEX = re.compile(r"^(#+)(\s.*)")
 # Regex to match TOC markers.
 TOC_REGEX = re.compile(r"<!--\s*toc\s*-->")
+# Code block boundary regex
```
sonniki, line 29: Comments must always end with `.` as per our conventions

linters/amp_check_md_toc_headers.py
```diff
@@ -26,6 +26,30 @@
...
+        >>> _is_code_block_boundary("```")
+        True
+        >>> _is_code_block_boundary("```python")
+        True
+        >>> _is_code_block_boundary("``` python")
+        False
+        >>> _is_code_block_boundary("  ```python")
+        True
+        >>> _is_code_block_boundary("Some text")
+        False
```
sonniki, line 50: This should come before the param/return descriptions

linters/amp_check_md_toc_headers.py
```diff
@@ -26,6 +26,30 @@
...
+    """
+    return CODE_BOUNDARY_REGEX.match(line.strip()) and len(line.split()) == 1
```
sonniki, line 52: For readability, let's split this line into steps, assign the final outcome to a variable and then return this variable

linters/amp_check_md_toc_headers.py
```diff
@@ -41,10 +65,15 @@ def fix_md_headers(lines: List[str], file_name: str) -> List[str]:
     """
     fixed_lines = []
     last_header_level = 0
+    inside_code_block = False
     for idx, line in enumerate(lines):
         fixed_line = line
+        # Toggle the inside_code_block flag if we find a code block boundary.
```
sonniki, line 71: We prefer not to use variable names in comments, because if the variable is renamed, we don't want to also have to change the comments.

linters/amp_check_md_toc_headers.py
```diff
@@ -41,10 +65,15 @@ def fix_md_headers(lines: List[str], file_name: str) -> List[str]:
...
+        if _is_code_block_boundary(line):
+            inside_code_block = not inside_code_block
```
sonniki, line 73: Actually, there is already a function `hstring.get_code_block_line_indices()` which would give you the indices of all lines that are inside code blocks. Could we use that instead of creating and using a new function here?

linters/test/test_amp_check_md_toc_headers.py
```diff
@@ -56,13 +56,16 @@ def test2(self) -> None:
         lines = hio.from_file(file_path).splitlines()
         updated_lines = lacmtohe.fix_md_headers(lines, file_path)
         # Check.
-        output = "\n".join(
-            ["# linter warnings", ""]
-            + []
-            + ["", "# linted file", ""]
-            + updated_lines
-        )
-        self.check_string(output, purify_text=True)
```
sonniki, line 65: `self.check_string()` uses a file in the `output` dir to store the expected output. Moving it to this function is fine, but then you should also delete the dir of this test in `linters/test/outcomes` since it's not used anymore.

linters/test/test_amp_check_md_toc_headers.py
```diff
@@ -91,6 +94,41 @@ def test3(self) -> None:
             updated_lines, txt_with_non_header_pound_signs.splitlines()
         )
 
+    def test4(self) -> None:
+        """
+        Test that python and bash code along with headers.
```
sonniki, line 99: -> Python

sonniki, line 99: Remove "that"

sonniki (changes requested):
Some of the comments regarding the new function would become obsolete if we switch to using the function from `hstring`. But please note the comments anyway since it would be useful for the future.

linters/amp_check_md_toc_headers.py
```diff
@@ -41,10 +42,11 @@ def fix_md_headers(lines: List[str], file_name: str) -> List[str]:
     """
     fixed_lines = []
     last_header_level = 0
+    code_line_indices = hstring.get_code_block_line_indices(lines)
```
sonniki, line 45: Let's add a comment here to explain what are we checking and why. Right now it's clear to us but in 6 months time it's not going to be

sonniki (changes requested):
Small thing for completeness, and then we'll merge

sonniki (approved):

Issue #553, opened 2025-04-13
Break in slow tests


gpsaggese:
GP Saggese

Friday at 5:43 PM
[@Sonya Nikiforova](https://causify.slack.com/team/U082BFT5S66)
 same glitch-y test linters/test/test_amp_dev_scripts.py::Test_linter_py1::test_linter_md1

That's a completely different error than before. Now it's failing because of some problem using the dockerized prettier. I'm guessing, it's because in [this PR](https://github.com/causify-ai/helpers/pull/544) the default behavior of dev_scripts_helpers/documentation/lint_notes.py was changed from using prettier installed on the host to using the dockerized executable.  To be more precise, the option to use prettier installed on the host was removed completely. BTW the PR was merged even though the slow tests were failing in it. Is there a way to completely block merging a PR if it breaks fast/slow tests? This is not the first time the build breaks because a PR was merged without accounting for the fact that it doesn't pass the fast/slow test checks.

FYI @dremdem @sonniki

dremdem:
> BTW the PR was merged even though the slow tests were failing in it. Is there a way to completely block merging a PR if it breaks fast/slow tests?

I think there are some settings already in place:
<img width="1408" alt="Image" src="https://github.com/user-attachments/assets/c043648b-b014-4bf9-9241-8cfb5dc0c6b8" />

It's possible they are already configured, but there might be a `bypass` with your username: 
<img width="1836" alt="Image" src="https://github.com/user-attachments/assets/eaa6bba6-8615-451a-8e18-2017cc4fe70e" />

heanhsok:
> That's a completely different error than before. Now it's failing because of some problem using the dockerized prettier. I'm guessing, it's because in https://github.com/causify-ai/helpers/pull/544 the default behavior of dev_scripts_helpers/documentation/lint_notes.py was changed from using prettier installed on the host to using the dockerized executable. To be more precise, the option to use prettier installed on the host was removed completely.

This PR https://github.com/causify-ai/helpers/pull/558 should fix the slow test

1) The issue is related to the mounting path of the source dir in the dockerized executables. It looks like we can't use the sibling container approach here. The children container approach should work. We can use the `hserver.use_docker_sibling_containers()` to check if sibling container is supported.
2) I have enabled this dockerized executables tests on our CI as it doesn't take too long to run. LMK if we should disable it like the others.

FYI @gpsaggese @sonniki 

```
[...]

PR #558, opened 2025-04-14
HelpersTask553_Break_in_slow_tests___dockerized_executables

heanhsok:
Task #553

- The issue is related to the mounting path of the source dir in the dockerized executables. It looks like we can't use the sibling container approach here. The children container approach should work. We can use the `hserver.use_docker_sibling_containers()` to check if sibling container is supported.

### DIFF
--- a/dev_scripts_helpers/documentation/lint_notes.py
+++ b/dev_scripts_helpers/documentation/lint_notes.py
@@ -3,7 +3,9 @@
 """
 Lint "notes" files.
 
-> lint_notes.py -i foo.txt -o bar.txt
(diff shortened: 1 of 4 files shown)


sonniki (approved):
BTW nevermind the Linter check failure, it's because of the bug that is being fixed in https://github.com/causify-ai/helpers/pull/535

samarth9008 (commented):
LG. Will this resolve other failures related to dockerized executables?

gpsaggese:
Let me merge this with some revert and merge to fix the break. Thanks @heanhsok. Good job understanding the issue. It wasn't easy.

Issue #541, opened 2025-04-11
Add skip files option to Linter


sonniki:
Sometimes we want to, for example, lint all files in the previous commit but explicitly skip and not lint some of those files. For this, we are missing an option like `skip_files`. Let's add that to [`i lint`](https://github.com/causify-ai/helpers/blob/4edbda1ec2aa9e1a15f0258b1c4418351790edc2/helpers/lib_tasks_lint.py#L181) and downstream (in the [base Linter script](https://github.com/causify-ai/helpers/blob/master/linters/base.py)). 

The interface should be the same as the `files` option (i.e. accepts a list of file paths separated by whitespace).

FYI @gpsaggese

sonniki:
Done

PR #572, opened 2025-04-16
HelpersTask541_Add_skip_files_option_to_Linter

aangelo9:
Related to #541.

Implements `--skip_files` option to `base.py` which accepts a list of file paths separated by whitespace. The input `skip_files` is first checked for validity then removed from the list of file paths to be lint.

Update `lint()` task in `helpers/lib_tasks_lint.py` to accept `--skip_files` args. Additional assertion is needed to ensure that the `--skip_files` option must be used with only one among `--modified`, `--branch`, `--last-commit`.


sonniki:
@aangelo9 is this meant to be a draft?

aangelo9:
I have implemented the necessary files but was not able to lint due to #578, so reverted it back to a draft because checks weren't all green.

sonniki:
> I have implemented the necessary files but was not able to lint due to https://github.com/causify-ai/helpers/issues/578, so reverted it back to a draft because checks weren't all green.

No need to turn into draft because of that. Turning into draft means "it's still work in progress, don't review, don't merge". Besides, even if Linter check is red, we can still see if other checks pass, especially the more important ones -- fast and slow tests.

aangelo9:
Understood. I’ll keep that in mind going forward.

linters/base.py
```diff
@@ -444,6 +453,12 @@ def _parse() -> argparse.ArgumentParser:
         action="store_true",
         help="Select files modified in the current branch with respect to master",
     )
+    parser.add_argument(
+        "--skip_files",
+        nargs="+",
+        type=str,
+        help="Files to skip from linting"
```
sonniki, line 460: "from" -> "during"

linters/base.py
```diff
@@ -102,6 +107,7 @@ def _get_files_to_lint(args: argparse.Namespace) -> List[str]:
     :return: paths of the files to lint
     """
     file_paths: List[str] = []
+    file_paths_to_skip: List[str] = []
```
sonniki, line 110: This can be moved closer to where it's used, for clarity. E.g.
```
file_paths_to_skip: List[str] = []
if args.skip_files:
    # Get the files that should be skipped during linting.
    file_paths_to_skip = args.skip_files
```

linters/base.py
```diff
@@ -64,19 +64,22 @@
 # #############################################################################
 
 
-def _filter_files(file_paths: List[str]) -> List[str]:
+def _filter_files(file_paths: List[str], *, file_paths_to_skip: Optional[List[str]] = None) -> List[str]:
```
sonniki, line 67: IMO `file_paths_to_skip` shouldn't be an optional parameter -- it's always there, and if `skip_files` flag is not present, then `file_paths_to_skip` is just an empty list. Correct?

linters/base.py
```diff
@@ -64,19 +64,22 @@
...
     """
-    Filter the list of files to be linted.
+    Filter the list of files by removing invalid or excluded files.
```
sonniki, line 69: -> "...invalid of excluded ones."

linters/base.py
```diff
@@ -87,6 +90,8 @@ def _filter_files(file_paths: List[str]) -> List[str]:
         is_valid &= ".ipynb_checkpoints/" not in file_path
         # Skip input and output files used in unit tests.
         is_valid &= not liutils.is_test_input_output_file(file_path)
+        # Skip explicitly specified files.
```
sonniki, line 93: `Skip files explicitly excluded by user.`

sonniki (changes requested):

helpers/lib_tasks_lint.py
```diff
@@ -200,6 +201,9 @@ def lint(  # type: ignore
     # To lint specific files:
     > i lint --files="dir1/file1.py dir2/file2.md"
 
+    # To lint the files changed in the last commit, excluding specific files:
+    > i lint --last_commit --skip_files="dir1/file1.py dir2/file2.md"
```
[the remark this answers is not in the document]
gpsaggese: Correct. Unfortunately, invoke uses a different convention for the cmd line options. Super annoying

helpers/lib_tasks_lint.py
```diff
@@ -229,13 +234,26 @@ def lint(  # type: ignore
     hlitauti.report_task()
     # Prepare the command line.
     lint_cmd_opts = []
-    # Add the file selection argument.
-    hdbg.dassert_eq(
+    # Enumerate file flags for assertion
```
gpsaggese, line 237: -> Count the command line options used to select files.

Also need a `.` at the end

helpers/lib_tasks_lint.py
```diff
@@ -229,13 +234,26 @@ def lint(  # type: ignore
...
+    )
+    # Add the file selection argument when using `--skip_files`.
+    hdbg.dassert_imply(
+        int(len(skip_files) > 0),
+        int(file_flags_set > 0),
+        msg=(
+            "--skip_files requires a file selection flag, specify exactly one among --files, "
+            "--dir-name, --modified, --last-commit, or --branch"
+        ),
+    )
```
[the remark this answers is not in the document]
gpsaggese: I agree with sonniki and with logic `(A => B) => (A \land C => B)`

linters/base.py
gpsaggese: Agree. Simpler APIs are always better.

gpsaggese (commented):
Nothing else besides @sonniki 's thorough review.

helpers/lib_tasks_lint.py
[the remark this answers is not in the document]
aangelo9: We could add `--last_commit` as an option so both could call the arg.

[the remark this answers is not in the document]
sonniki: Maybe there is a misunderstanding. Me and GP are just talking about the way the flags should be written when the invoke is called: with a hyphen instead of an underscore

helpers/lib_tasks_lint.py
```diff
@@ -200,6 +201,9 @@ def lint(  # type: ignore
     # To lint specific files:
     > i lint --files="dir1/file1.py dir2/file2.md"
 
+    # To lint the files changed in the last commit, excluding specific files:
+    > i lint --last-commit --skip_files="dir1/file1.py dir2/file2.md"
```
sonniki, line 205: -> `--skip-files`

helpers/lib_tasks_lint.py
```diff
@@ -213,6 +217,7 @@ def lint(  # type: ignore
     :param stage: the image stage to use (e.g., "prod", "dev", "local")
     :param version: the version of the container to use
     :param files: specific files to lint (e.g. "dir1/file1.py dir2/file2.md")
+    :param skip_files: specific files to skip during lint (e.g. "dir1/file1.py dir2/file2.md")
```
sonniki, line 220: -> "during linting"

linters/base.py
```diff
@@ -127,8 +131,12 @@ def _get_files_to_lint(args: argparse.Namespace) -> List[str]:
         cmd = f"find {dir_name} -name '*' -type f"
         _, output = hsystem.system_to_string(cmd)
         file_paths = output.split("\n")
+    file_paths_to_skip: List[str] = []
+    if args.skip_files:
+        # Get and validate the files to skip that were explicitly specified.
```
sonniki, line 136: -> "Get the files to skip during linting."

sonniki (changes requested):
Minor things and then we'll merge

sonniki (approved):

Issue #280, opened 2025-02-06
Markdown code blocks are reflowed


sonniki:
1) 
```markdown
    [tool.poetry.dependencies]
    ...
    pytest-timeout = "*"
    ...
```

turns into 

```markdown
    [tool.poetry.dependencies] ... pytest-timeout = "\*" ...
```

2)
```markdown
  amp==1.1.4
  async_solipsism==0.3
  beautifulsoup4==4.11.1
  botocore==1.24.37
  cvxopt==1.3.0
  cvxpy==1.2.0
  dill==0.3.4
  environs==9.5.0
  ...
```

turns into 

```markdown
    amp==1.1.4 async_solipsism==0.3 beautifulsoup4==4.11.1 botocore==1.24.37
  cvxopt==1.3.0 cvxpy==1.2.0 dill==0.3.4 environs==9.5.0 ...
```

See `docs/work_tools/all.devops_docker.how_to_guide.md` after linting

sonniki:
@sandeepthalapanane this is a Linter bug fix issue.
Some useful links for debugging Linter:
- Entrypoint of the [`i lint`](https://github.com/causify-ai/helpers/blob/7845d0a36fde3f3c2b4652b1a5e29eb8cd6b6cc6/helpers/lib_tasks_lint.py#L181) command
- [Main Linter script](https://github.com/causify-ai/helpers/blob/master/linters/base.py) that calls all the steps
- [Linter dev doc](https://github.com/causify-ai/helpers/blob/master/docs/coding/all.developing_linter.how_to_guide.md)

sonniki:
The usual flow is
1) Create unit tests to reproduce the problem
2) Fix the bug
3) Confirm with unit tests from step (1) that the problem is fixed

sonniki:
This solution is fine if nothing else works, but just wondering, do we know why it's happening with these code blocks but not the others?

gpsaggese:
Let's create examples to make sure we understand what is the problem. Adding the ignore is not a bad idea.

sandeepthalapanane:
I attempted to lint the below example with and without the markdown code block. It seems to be treating the markdown code fencing as another markdown, and the text inside the fence block is being formatted as a paragraph. The linter script passes `--prose-wrap` argument to `Prettier`. It reflows any lines that are not fenced code or bullet lists into a single wrapped paragraph. When markdown is not used, `Prettier` is considering it as a fenced block and stops formatting. 

```markdown
```markdown
[...]

PR #573, opened 2025-04-16
HelpersTask280_Markdown_code_blocks_are_reflowed

sandeepthalapanane:
Addressing #280 

I changed the markdown block to a text block and also added a unit test with a text fenced block. As of now, I have made sure the test is skipped (#553), but I ran the test locally on my machine, and it passed.


sandeepthalapanane:
If I lint the file `linters/test/test_amp_dev_scripts.py`, it modifies the unit test input python scripts. So, I chose not to run the linter.

linters/test/test_amp_dev_scripts.py
```diff
@@ -126,6 +127,31 @@ def test_linter_md1(self) -> None:
         # Check.
         self.check_string(output, purify_text=True)
 
+    @pytest.mark.slow("About 6 sec")
```
sonniki, line 130: Could you add the same TODO as for `test_linter_md1()` for @heanhsok to remove the skip when the issue is resolved?

sonniki (changes requested):
One TODO to add and then we'll merge

PR #344, opened 2025-03-07
HelpersTask339_Develop_the_docker_release_multi_arch_prod_image_invoke

dremdem:
#339

- docker_release_multi_arch_prod_image() invoke

Issue #339, opened 2025-03-04
Develop the `docker_release_multi_arch_prod_image()` invoke


dremdem:
Follow-up from the [#11418](https://github.com/causify-ai/helpers/blob/master/helpers/lib_tasks_docker_release.py#L541)

https://github.com/causify-ai/helpers/pull/324#discussion_r1977862375

# Description

Add an end-to-end function for releasing multi-arch prod image. Something like this
[docker_release_prod_image](https://github.com/causify-ai/helpers/blob/d686523eb8c5f6784d4a96e1856c03c0c3c84fa9/helpers/lib_tasks_docker_release.py#L878C5-L878C30)

# Steps to implement 

- Analyze `docker_release_prod_image()` and `docker_release_multi_build_dev_image()` and develop the new `docker_release_multi_arch_prod_image()`
- Change the docs if needed

dremdem:
FYI
@samarth9008

PR #586, opened 2025-04-17
HelpersTask339_Develop_the_docker_release_multi_arch_prod_image_invoke_2

dremdem:
#339 

moved from https://github.com/causify-ai/helpers/pull/344

- docker_release_multi_arch_prod_image() invoke


helpers/lib_tasks_docker_release.py
```diff
@@ -930,6 +930,81 @@ def docker_release_prod_image(  # type: ignore
     _LOG.info("==> SUCCESS <==")
 
 
+@task(iterable=["docker_registries"])
```
gpsaggese, line 933: -> docker_registry

helpers/lib_tasks_docker_release.py
```diff
@@ -930,6 +930,81 @@ def docker_release_prod_image(  # type: ignore
...
+    skip_tests=False,
+    fast_tests=True,
+    slow_tests=True,
+    superslow_tests=False,
+    qa_tests=True,
+    docker_registries=None,
+    container_dir_name=".",
+):
+    """
+    Build, test, and release to Docker registries the multi-aech prod image.
```
gpsaggese, line 947: multi-arch

helpers/lib_tasks_docker_release.py
dremdem: Changed.

helpers/lib_tasks_docker_release.py
dremdem: Fixed.

gpsaggese (commented):
LGTM

helpers/lib_tasks_docker_release.py
```diff
@@ -930,6 +930,81 @@ def docker_release_prod_image(  # type: ignore
...
+            "No Docker registries provided, using default: %s", docker_registry
+        )
+    # 1) Build prod image.
+    docker_build_multi_arch_prod_image(
+        ctx,
+        version,
+        cache=cache,
+        container_dir_name=container_dir_name,
+        multi_arch="linux/amd64,linux/arm64",
+    )
```
samarth9008, line 979: I think we are missing few params

```
def docker_build_multi_arch_prod_image(  # type: ignore
    ctx,
    version,
    cache=True,
    base_image="",
    user_tag="",
    container_dir_name=".",
    tag=None,
    multi_arch="linux/amd64,linux/arm64",
):
```

helpers/lib_tasks_docker_release.py
```diff
@@ -930,6 +930,81 @@ def docker_release_prod_image(  # type: ignore
...
+    if slow_tests:
+        hlitapyt.run_slow_tests(ctx, stage=stage, version=version)
+    if superslow_tests:
+        hlitapyt.run_superslow_tests(ctx, stage=stage, version=version)
+    # 3) Run QA tests.
+    if qa_tests:
+        hlitapyt.run_qa_tests(ctx, stage=stage, version=version)
+    # 4) Push prod image.
+    for registry in docker_registry:
+        docker_tag_push_multi_arch_prod_image(
```
samarth9008, line 999: Same with this function call

samarth9008 (commented):
Some comments.

BTW did we test if the flow is working fine? LMK how you did it since we don't wanna override a broken image. Although it won't affect the tests since this is a prod image but can break linter

helpers/lib_tasks_docker_release.py
dremdem: Hmm, I'm not sure I got the point.

I'm not passing the following arguments, which already have default values:
```
base_image = "",
user_tag = "",
tag = None,
```

helpers/lib_tasks_docker_release.py
dremdem: The same as above. 
I missed the `base_image` only that is "" by default.

dremdem:
> BTW did we test if the flow is working fine? LMK how you did it since we don't wanna override a broken image. Although it won't affect the tests since this is a prod image but can break linter

IMO these invoke tasks just like a guide how to release the image. 
It's really hard to run it end-2-end in a real life, since the process is time consuming and kind of unpredictable. 
Speaking about the testing, yes, I tested it like: 
- comment out all the calls the functions
- add the `_LOG.info()` instead of them
- called this invoke target as: 
```
i docker_release_multi_arch_prod_image --version 1.2.0
```

```
invoke docker_release_multi_arch_prod_image \
    --version 1.2.0 \
    --docker-registries dockerhub.causify \
    --docker-registries aws_ecr.ck
```

```
invoke docker_release_multi_arch_prod_image \
    --version 1.2.0 \
    --docker-registries dockerhub.causify \
    --docker-registries aws_ecr.ck \
    --skip-tests
```
- and check results

gpsaggese:
I know what you mean by "kind of long and unpredictable". My take is that if we had a CD approach and run the tests all the times then it would be easier to keep the pipeline going.
For now, we end up running them in chunks.

Let's document this in the md and so we can move on.

gpsaggese (commented):
LGTM

gpsaggese (commented):
LGTM. Can we merge?

helpers/lib_tasks_docker_release.py
samarth9008: What if user wants to pass some value to it?

dremdem:
@samarth9008 
Can we push it forward since it is blocking also: 
https://github.com/causify-ai/helpers/pull/588

samarth9008:
Feel free to merge if you think its working fine. I'll review later on

Issue #515, opened 2025-04-08
Replace cryptokaizen with causify-ai in links


sonniki:
We still have links in the codebase that use the old GH Org name `cryptokaizen` instead of the new one `causify-ai`. Let's find all instances and replace with the new name.

PR #613, opened 2025-04-24
HelpersTask515-replace-cryptokaizen-with-causify-ai-links

neomisule:
Link to Issue: #515 

Replacing cryptokaizen with causify-ai, ck with csfy and _ck with _csfy.

FYI @sonniki @samarth9008 @gpsaggese


neomisule:
I was thinking of adding a shell script to automatically replace all instances of `ck` with `csfy`, but I'm concerned it might unintentionally change some variable or function names that include `ck`, such as `back` to `bacsfy`, which could lead to bugs or confusion. Do you think it's better to handle those changes manually?

Also, the current links that reference `cryptokaizen` are already broken. I tried replacing `cryptokaizen` with `causify_ai` in the URLs, but they still don't work. Even the S3 bucket links include `cryptokaizen` in the path, so should we be replacing it with `causify-ai` across all those places, or is there a different approach we should take?

@sonniki @samarth9008

sonniki:
Replacing manually and only in links to GH pages would be enough here

Issue #619, opened 2025-04-25
Remove header level difference assertion


sonniki:
In `check_header_list()` in `helpers/hamarkdown.py`, there is a check that asserts if consecutive elements in the header list differ by more than 1 level, see [here](https://github.com/causify-ai/helpers/blob/5ff3dc086faa689a2635494f906a7f86e1f100e4/helpers/hmarkdown.py#L407C18-L407C67).

However, this is a completely normal situation, e.g.:

```
# First-level header
## Second-level header A
### Third-level header
#### Fourth-level header
## Second-level header B
```
There is no requirement that after introducing a fourth-level header we need to go back to the third-level header. We should be able to go directly to the second-level header.
In our docs, this situation is present, for example, in [`docs/work_tools/all.development.how_to_guide.md`](https://github.com/causify-ai/helpers/blob/5ff3dc086faa689a2635494f906a7f86e1f100e4/docs/work_tools/all.development.how_to_guide.md?plain=1#L274)

```
- [Publish a notebook](#publish-a-notebook)
  * [Detailed instructions](#detailed-instructions)
  * [Publish notebooks](#publish-notebooks)
  * [Open a published notebook](#open-a-published-notebook)
    + [Start a server](#start-a-server)
    + [Using the dev box](#using-the-dev-box)
    + [Using Windows browser](#using-windows-browser)
- [How to create a private fork](#how-to-create-a-private-fork)
```
From the fourth-level `Using Windows browser` we go straight to the second-level `How to create a private fork`

@gpsaggese according to `git blame` it was your code, right? WDYT?

gpsaggese:
Agreed that the assertion is not correct.
I'd say when it increments, it should increment by 1. When it goes down, it can jump.

It should be an easy fix. GFI?

sonniki:
Done

PR #620, opened 2025-04-26
HelpersTask619_Remove_header_level_difference_assertion

sandeepthalapanane:
Addressing #619 

Modified the header level validation to allow decreasing by multiple levels but still require increasing by only one level at a time. 

Should we add a test case to verify this?

### DIFF
--- a/helpers/hmarkdown.py
+++ b/helpers/hmarkdown.py
@@ -404,15 +404,15 @@ def check_header_list(header_list: HeaderList) -> None:
             header_list[0].line_number,
             header_list[0].level,
         )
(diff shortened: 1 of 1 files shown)


sonniki:
> Should we add a test case to verify this?

The answer to this is _always_ "YES". You don't have to ask

sandeepthalapanane:
@sonniki I implemented the test cases, but linting `test_hmarkdown.py` modifies the content of the test cases. So, I skipped linting the test case script.

sonniki (changes requested):

sonniki (approved):

gpsaggese (commented):
LG

Issue #618, opened 2025-04-25
Links formatted inside quotation marks


sonniki:
By `linters/amp_fix_md_links.py`, bare links turn into Markdown-style links in contexts like this:

```
<img width="505" alt="" src="https://github.com/kaizen-ai/kaizenflow/assets/31514660/69fbabec-300c-4f7c-94fc-45c5da5a6817">
```

Links should remain as-is if they are located inside quotation marks.

To reproduce: lint `docs/coding/all.submit_code_for_review.how_to_guide.md`.

gpsaggese:
Is it a GSI? If so, pls add the label. (Let's always be thinking about "what can be outsourced", otherwise it's difficult to find after the fact)

sonniki:
Done

PR #621, opened 2025-04-26
HelpersTask618_Links_formatted_inside_quotation_marks

sandeepthalapanane:
Addressing #618 

Modified the bare link regex pattern to catch URLs inside quotation marks. A test case was added to verify if the modified pattern works.

Note - The modified regex pattern does not match URLs inside single quotes and URLs inside quotation marks starting with text. E.g. `'https://example.com/path'`, `"Word https://example.com/path"`. 

I did a grep and checked if there are any such instances right now in the repo, but there aren't any. If necessary, we can make it robust to the above examples by using a complex regex pattern or a helper function.


sonniki (approved):

sonniki:
@sandeepthalapanane pls don't forget to put yourself into the "Assignees" field in PRs -- we will use this field in some of the automatic metrics that are under development

Issue #615, opened 2025-04-24
Come up with a testing approach for release flow


gpsaggese:
As per https://github.com/causify-ai/helpers/pull/590#pullrequestreview-2792676001 when we improve / refactor critical code like the docker flow we don't have a good way to test it, besides running it in production.

We are going to have soon a way to run tests outside a container, which will make it possible to run tests for this part of the repo. The next problem is how to test without actually running?
An idea is to check that, under certain mocking, the sequence of system calls generated is the expected one.

FYI @heanhsok @dremdem @samarth9008

sandeepthalapanane:
As per your suggestions above, we could use `unittest.mock` to test the Docker release workflow without requiring actual Docker operations and to mock external dependencies

### Code Structure

1. **Mock Infrastructure**
   - Mock Docker client interactions
   - Mock system calls (`hsystem.system`, `hlitauti.run`)
   - Mock registry client for AWS ECR and DockerHub
   - Mock environment variables

2. **Test Categories**

   a. **Command Generation Tests**
   - Check that the correct sequence of commands is generated for different scenarios
   - Test multi-architecture build commands
   - Validate registry-specific commands

   b. **Error Handling Tests**
   - Test unsupported registry scenarios
   - Verify version validation
   - Test system error propagation

gpsaggese:
The plan looks reasonable. I think you understood the issue.

The goal is to ensure that by mocking everything that needs to be mocked, we can get a sequence of commands that is the one expected.
After testing end-to-end with the mocks, we can unit test "deterministic" chunks of code by extracting code that doesn't interact and just testing it in isolation.

My proposal is to test a single function, showing how things will work and after that PR start expanding the coverage as we go.

@heanhsok, any thoughts?

sandeepthalapanane:
> The plan looks reasonable. I think you understood the issue.
> 
> The goal is to ensure that by mocking everything that needs to be mocked, we can get a sequence of commands that is the one expected. After testing end-to-end with the mocks, we can unit test "deterministic" chunks of code by extracting code that doesn't interact and just testing it in isolation.
> 
> My proposal is to test a single function, showing how things will work and after that PR start expanding the coverage as we go.
> 
> [@heanhsok](https://github.com/heanhsok), any thoughts?

I think I will start by testing the `docker_build_local_image` (both single arch and multi-arch) since it's the first step in the docker flow and covers many things.

heanhsok:
[...]

PR #640, opened 2025-04-29

sandeepthalapanane:
Added unit tests for `docker_build_local_image` (both single arch and multi-arch) (#615)

- Added unittest with mocks to verify Docker build command sequences
- Tests both single and multi-architecture build flows
- Includes assertion error messages with actual and expected command comparison for debugging

### DIFF
--- a/helpers/test/test_lib_tasks_docker_release.py
+++ b/helpers/test/test_lib_tasks_docker_release.py
@@ -1,12 +1,48 @@
 import logging
 import re
 import unittest.mock as umock
+from typing import List, Tuple
 
 import helpers.hsystem as hsystem
 import helpers.hunit_test as hunitest
 import helpers.lib_tasks_docker_release as hltadore
 
 _LOG = logging.getLogger(__name__)
+
+
+def normalize_command(cmd: str) -> str:
+    """
+    Normalize a command string by removing line continuations and spaces around
+    special characters.
+
+    :param cmd: command string to normalize
+    :return: normalized command string
+    """
+    cmd = re.sub(r"\\\s*\n\s*", " ", cmd)
+    chars_to_normalize = ["=", ",", "|"]
+    escaped_chars = "".join(re.escape(ch) for ch in chars_to_normalize)
+    cmd = re.sub(rf"\s*([{escaped_chars}])\s*", r"\1", cmd)
+    return cmd
+
+
+def convert_commands_to_strings(
+    actual_cmds: List[str], expected_cmds: List[str]
+) -> Tuple[str, str]:
+    """
+    Convert a list of commands to strings and normalize them.
+
+    :param actual_cmds: list of actual command strings
+    :param expected_cmds: list of expected command strings
+    :return: normalized actual commands string and normalized expected
+        commands string
+    """
+    # Normalize each command in both lists.
+    actual_normalized = [normalize_command(cmd) for cmd in actual_cmds]
+    expected_normalized = [normalize_command(cmd) for cmd in expected_cmds]
+    # Convert to strings.
+    actual_str = "\n".join(actual_normalized)
+    expected_str = "\n".join(expected_normalized)
+    return actual_str, expected_str
 
 
 # #############################################################################
@@ -20,6 +56,7 @@ class TestDockerBuildLocalImage1(hunitest.TestCase):
         """
         Set up test environment and initialize all necessary mocks.
         """
+        super().setUp()
         # Mock system calls.
         self.system_patcher = umock.patch("helpers.hsystem.system")
         self.mock_system = self.system_patcher.start()
@@ -36,6 +73,7 @@ class TestDockerBuildLocalImage1(hunitest.TestCase):
             "helpers.lib_tasks_docker.docker_login"
         )
         self.mock_docker_login = self.docker_login_patcher.start()
+        #
         self.user = hsystem.get_user_name()
 
     def tearDown(self) -> None:
@@ -46,6 +84,7 @@ class TestDockerBuildLocalImage1(hunitest.TestCase):
         self.run_patcher.stop()
         self.version_patcher.stop()
         self.docker_login_patcher.stop()
+        super().tearDown()
 
     def test_docker_build_single_arch(self) -> None:
         """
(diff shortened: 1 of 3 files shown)


helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -0,0 +1,175 @@
...
+        generated for building a local Docker image with single
+        architecture.
+        """
+        # Prepare inputs.
+        mock_ctx = umock.MagicMock()
+        test_version = "1.0.0"
+        test_base_image = "test-registry.com/test-image"
+        test_multi_arch = ""
+        # Call tested function.
+        hltadore.docker_build_local_image.body(
```
gpsaggese, line 64: What does `.body` do? Can you add a comment?

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -0,0 +1,175 @@
...
+            ctx=mock_ctx,
```
gpsaggese, line 65: we use `param=...` only for optional params, so you shouldn't need them for a few of them

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -0,0 +1,175 @@
...
+            version=test_version,
+            base_image=test_base_image,
+            multi_arch=test_multi_arch,
+            poetry_mode="update",
+            cache=False,
+        )
+        # Extract the command arguments from the call.
+        actual_cmds = [call[0][1] for call in self.mock_run.call_args_list]
+        # Check output.
+        expected_cmds = [
```
gpsaggese, line 75: Let's just use a str so it's easier to compare.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -0,0 +1,175 @@
...
+            f"--version {test_version} "
+            f"--cmd 'cp -f /install/poetry.lock.out /install/pip_list.txt .' --skip-pull",
+            "cp -f poetry.lock.out ./devops/docker_build/poetry.lock",
+            "cp -f pip_list.txt ./devops/docker_build/pip_list.txt",
+            f"docker image ls {test_base_image}:local-{self.user}-1.0.0",
+        ]
+        # Normalize both expected and actual commands.
+        expected = [self._normalize_command(cmd) for cmd in expected_cmds]
+        actual = [self._normalize_command(cmd) for cmd in actual_cmds]
+        self.assertEqual(
```
gpsaggese, line 96: Better to use assert_equal from our framework since it has a lot of interesting features.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -0,0 +1,175 @@
...
+            expected,
+            actual,
+            f"Expected commands: {expected}\nActual commands: {actual}",
+        )
+
+    def _normalize_command(self, cmd: str) -> str:
```
gpsaggese, line 157: FYI assert_equal has a lot of options including fuzzy to do fuzzy matching.

You might still need some of normalize

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -0,0 +1,175 @@
...
+            " --file devops/docker_build/dev.Dockerfile -",
+            "invoke docker_cmd  "
+            "--stage local "
+            f"--version {test_version} "
+            f"--cmd 'cp -f /install/poetry.lock.out /install/pip_list.txt .' --skip-pull",
+            "cp -f poetry.lock.out ./devops/docker_build/poetry.lock",
+            "cp -f pip_list.txt ./devops/docker_build/pip_list.txt",
+            f"docker image ls {test_base_image}:local-{self.user}-1.0.0",
+        ]
+        # Normalize both expected and actual commands.
```
gpsaggese, line 93: You can factor out some of this common code across tests

gpsaggese (commented):
Good work. Some nits and merge

helpers/test/test_lib_tasks_docker_release.py
sandeepthalapanane: I factored out the common code outside the test class, to make it usable for other tests.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -0,0 +1,200 @@
...
+        # parsing and run the raw implementation.
+        hltadore.docker_build_local_image.body(
+            mock_ctx,
+            test_version,
+            cache=False,
+            base_image=test_base_image,
+            poetry_mode="update",
+        )
+        # Extract the command arguments from the call.
+        actual_cmds = [call[0][1] for call in self.mock_run.call_args_list]
```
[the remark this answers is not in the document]
sandeepthalapanane: I will move the extraction out of the test class and create a helper function since it will also be useful for other tests.

helpers/test/test_lib_tasks_docker_release.py
dremdem: Can you pls explain here. No need to change the comment atm. 
Why we need to run it like this? 
I wondering also since we never use it before. 
For example: 
https://github.com/causify-ai/helpers/blob/aec44797d06dc7a4f80b1b959dd8d7db09148d58/helpers/test/test_lib_tasks.py#L486

sandeepthalapanane: I don't think we need to use `.body`. The issue arose because I was passing the mandatory `ctx` argument as `ctx=mock_ctx` instead of just passing `mock_ctx`.

dremdem (commented):
It’s looking better with each step.
Just a few comments.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -0,0 +1,200 @@
...
+            " --build-arg CLEAN_UP_INSTALLATION=True "
+            f" --tag {test_base_image}:local-{self.user}-1.0.0 "
+            " --file devops/docker_build/dev.Dockerfile -",
+            "invoke docker_cmd  "
+            "--stage local "
+            f"--version {test_version} "
+            f"--cmd 'cp -f /install/poetry.lock.out /install/pip_list.txt .' --skip-pull",
+            "cp -f poetry.lock.out ./devops/docker_build/poetry.lock",
+            "cp -f pip_list.txt ./devops/docker_build/pip_list.txt",
+            f"docker image ls {test_base_image}:local-{self.user}-1.0.0",
```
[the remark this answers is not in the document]
sandeepthalapanane: I had to revert the change as per @gpsaggese's new commit.

sandeepthalapanane:
@dremdem @gpsaggese Would you like me to add the same test for `docker_build_prod_image` in this pull request? This way, we can cover all image build functions in one PR.

dremdem:
> @dremdem @gpsaggese Would you like me to add the same test for `docker_build_prod_image` in this pull request? This way, we can cover all image build functions in one PR.

IMO, it's better to keep it as is, since it easer to review and merge.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -0,0 +1,184 @@
...
+        docker image ls test-registry.com/test-image:local-$USER_NAME-1.0.0
+        """
+        # Check output.
+        self.assert_equal(
+            actual_cmds,
+            exp,
+            purify_text=True,
+            fuzzy_match=True,
+            remove_lead_trail_empty_lines=True,
+            dedent=True,
```
dremdem, line 183: I think it’s redundant since fuzzy_match=True.
Feel free to check and remove it everywhere if it works.

dremdem (approved):
LGTM
Feel free to ask Samarth for review after addressing one nit.

helpers/test/test_lib_tasks_docker_release.py
gpsaggese: It's ok, even if redundant. We'll keep working on this so let's merge

PR #583, opened 2025-04-16
HelpersTask578_Docker_fails_due_to_missing_Docker_socket

aangelo9:
Related to #578

Current Pipeline:
```
henv.env_to_str()
 └── henv.get_system_signature()
   └── hserver.get_docker_info()
     └── hserver.docker_needs_sudo()
       └── assert False # If Docker fails
```

We want to insert `hserver.is_external_dev()`.
Approach:

Insert in `hserver.get_docker_info()`:
```
def get_docker_info() -> str:
    ...
    if is_external_dev():
        docker_needs_sudo_ = None
    else:
        docker_needs_sudo_ = docker_needs_sudo()
    ...
```

Insert in `henv.get_system_signature()`:
```
def get_system_signature(git_commit_type: str = "all") -> Tuple[str, int]:
    ...
    # Add Docker info conditionally
    if hserver.is_external_dev():
        txt_tmp = "External dev: skipping Docker check"
    else:
        txt_tmp = hserver.get_docker_info()
    ...
```

Implementation in `hserver.get_docker_info()` might be better for future use cases if its reused elsewhere. It also encapsulates safety logic in one place since it already calls `is_inside_docker_`.

Issue #629, opened 2025-04-28
Add tests for dockerized executable sync_gh_issue_labels


sandeepthalapanane:
As per https://github.com/causify-ai/helpers/pull/600#issuecomment-2836945024, we need to test the dockerized executable [sync_gh_issue_labels.py](https://github.com/causify-ai/helpers/blob/master/dev_scripts_helpers/github/sync_gh_issue_labels.py). The initial idea is to mock GitHub interactions to verify the synchronization logic (similar to #619) and also add unittests for the helper functions. 

FYI @heanhsok @gpsaggese @sonniki

gpsaggese:
Adding to project

Issue #578, opened 2025-04-16
Docker fails due to missing Docker socket


aangelo9:
When trying to run invoke commands, i ran into:
```
# invoke print_env
12:02:00 - INFO  hdbg.py init_logger:1018                               > cmd='/venv/bin/invoke print_env'
12:02:00 - WARN  hserver.py _raise_invalid_host:777                     Don't recognize host: host_os_name=Linux, am_host_os_name=None
[sudo] password for ubuntu: 
sudo: a password is required
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?

Run 'docker run --help' for more information
[sudo] password for ubuntu: 
sudo: a password is required
Traceback (most recent call last):
  File "/venv/bin/invoke", line 8, in <module>
    sys.exit(program.run())
             ^^^^^^^^^^^^^
  File "/venv/lib/python3.12/site-packages/invoke/program.py", line 398, in run
    self.execute()
  File "/venv/lib/python3.12/site-packages/invoke/program.py", line 583, in execute
    executor.execute(*self.tasks)
  File "/venv/lib/python3.12/site-packages/invoke/executor.py", line 140, in execute
    result = call.task(*args, **call.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/venv/lib/python3.12/site-packages/invoke/tasks.py", line 138, in __call__
    result = self.body(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/helpers/lib_tasks_print.py", line 92, in print_env
    henv.env_to_str(
  File "/app/helpers/henv.py", line 543, in env_to_str
    msg += get_system_signature()[0] + "\n"
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/app/helpers/henv.py", line 505, in get_system_signature
    txt_tmp = hserver.get_docker_info()
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/helpers/hserver.py", line 683, in get_docker_info
    docker_needs_sudo_ = docker_needs_sudo()
                         ^^^^^^^^^^^^^^^^^^^
  File "/app/helpers/hserver.py", line 631, in docker_needs_sudo
    assert False, "Failed to run docker"
           ^^^^^
AssertionError: Failed to run docker
```

[...]

PR #649, opened 2025-05-02
HelpersTask629_Add_tests_for_dockerized_executable_sync_gh_issue_labels

sandeepthalapanane:
Addressing #629 

Implemented tests to check the Docker flow sequence for the Docker executable and also for the functionality class.


dev_scripts_helpers/github/test/test_dockerized_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,66 @@
...
+
+import yaml
+
+import dev_scripts_helpers.github.dockerized_sync_gh_issue_labels as dshgdsgil
+import helpers.hunit_test as hunitest
+
+_LOG = logging.getLogger(__name__)
+
+
+def _get_label_data() -> dict:
```
gpsaggese, line 13: -> Dict[str, str]

dev_scripts_helpers/github/test/test_dockerized_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,66 @@
...
+
+class Test_Label1(hunitest.TestCase):
+
+    def test_load_labels(self) -> None:
+        """
+        Test loading labels from a YAML file.
+        """
+        label_data = _get_label_data()
+        curr_dir = os.getcwd()
+        with tempfile.NamedTemporaryFile(
```
gpsaggese, line 30: you should use the self.get_scratch_space()

Can you make sure to read the unit test docs

./helpers_root/docs/coding/all.run_unit_tests.how_to_guide.md
./helpers_root/docs/coding/all.write_unit_tests.how_to_guide.md

dev_scripts_helpers/github/test/test_dockerized_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,66 @@
...
+    def test_save_labels(self) -> None:
+        """
+        Test saving labels to a YAML file.
+        """
+        label_data = _get_label_data()
+        label = dshgdsgil.Label(
+            label_data["name"], label_data["description"], label_data["color"]
+        )
+        curr_dir = os.getcwd()
+        with tempfile.NamedTemporaryFile(
```
gpsaggese, line 53: Same

dev_scripts_helpers/github/test/test_dockerized_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,66 @@
...
+            dir=curr_dir, mode="w", encoding="utf-8", delete=False
+        ) as tmp_file:
+            tmp_file_name = tmp_file.name
+        try:
+            dshgdsgil.Label.save_labels([label], tmp_file_name)
+            with open(tmp_file_name, "r", encoding="utf-8") as f:
```
gpsaggese, line 59: Why not reading them with the load_labels?

You can do a turn around test, create an object, save it, read it and then compare to the original one

dev_scripts_helpers/github/test/test_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,114 @@
...
+# #############################################################################
+
+
+class Test_sync_gh_issue_labels1(hunitest.TestCase):
+
+    def setUp(self) -> None:
+        """
+        Set up test fixtures.
+        """
+        super().setUp()
```
gpsaggese, line 22: Ugh. There is so much mocking that at best we test only what you already know it works, at worst we don't really test anything.

Our rule for mocking is to mock only the external services (in this case we should mock the interaction with GH)

Can we improve helpers_root/docs/coding/all.write_unit_tests.how_to_guide.md
explaining these, making examples and so on?

You can work with @heanhsok on this

dev_scripts_helpers/github/test/test_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,114 @@
...
+            "test-org",
+            "test-repo",
+            "GH_TOKEN",
+            dry_run=True,
+            no_interactive=True,
+            prune=True,
+            backup=True,
+        )
+        actual_cmd = self.mock_system.call_args[0][0]
+        expected_cmd = (
```
gpsaggese, line 99: let's make it a string, since the framework is designed for that

dev_scripts_helpers/github/test/test_sync_gh_issue_labels.py
sandeepthalapanane: I completely revised the test, using the reference from [dev_scripts_helpers/documentation/test/test_dockerized_pandoc.py](https://github.com/causify-ai/helpers/blob/master/dev_scripts_helpers/documentation/test/test_dockerized_pandoc.py). Unfortunately, I am unable to verify the testing because of an issue with the Docker daemon in my local Ubuntu system. I have added `pytest.skip` to skip the test.

dev_scripts_helpers/github/test/test_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        Test that the dockerized executable reads the input file, performs the
+        necessary operations, and writes the backup file in the git root
+        directory.
+        """
+        # Set up mock GitHub repo and labels.
+        mock_label = umock.Mock()
+        mock_label.name = "bug"
+        mock_label.color = "f29513"
+        mock_label.description = "Something isn't working"
```
heanhsok, line 39: Can we group the values that we are mocking into a dict or an obj? This way, it's easier to understand what we're mocking at a quick glance.

dev_scripts_helpers/github/test/test_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,71 @@
...
+        mock_repo = umock.Mock()
+        mock_repo.get_labels.return_value = [mock_label]
+        mock_client = umock.Mock()
+        mock_client.get_repo.return_value = mock_repo
+        mock_github.return_value = mock_client
+        # Prepare inputs.
+        in_dir_name = self.get_input_dir()
+        input_file_path = os.path.join(in_dir_name, "test_gh_issues_labels.yml")
+        owner = "test-org"
+        repo = "test-repo"
```
heanhsok, line 49: Same.

dev_scripts_helpers/github/test/test_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,71 @@
...
+        reason="Disabled because of CmampTask10710",
+    )
+    @umock.patch(
+        "dev_scripts_helpers.github.dockerized_sync_gh_issue_labels.github.Github"
+    )
+    def test1(self, mock_github: umock.Mock) -> None:
+        """
+        Test that the dockerized executable reads the input file, performs the
+        necessary operations, and writes the backup file in the git root
+        directory.
```
heanhsok, line 33: It looks like this is testing too many things. If the test fails, how do we know if it's because it has trouble reading the input file, or performing an operation (which one?), or because it's not able to save to the backup file?

dev_scripts_helpers/github/test/test_dockerized_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,49 @@
...
+            # Save the label.
+            dshgdsgil.Label.save_labels([input_label], tmp_file_name)
+            # Load the label.
+            output_labels = dshgdsgil.Label.load_labels(tmp_file_name)
+            self.assertEqual(len(output_labels), 1)
+            output_label = output_labels[0]
+            # Check.
+            self.assertEqual(output_label.name, input_label.name)
+            self.assertEqual(output_label.description, input_label.description)
+            self.assertEqual(output_label.color, input_label.color)
```
heanhsok, line 47: When there are multiple variables to compare like this, we usually prefer to do a string comparison in just one assert statement with a fuzzy match.
```
self.assert_equal(act, exp, fuzzy_match=False)
```

dev_scripts_helpers/github/test/test_dockerized_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,49 @@
...
+        """
+        Test saving and loading a label object.
+        """
+        # Prepare inputs.
+        label_data = _get_label_data()
+        input_label = dshgdsgil.Label(
+            label_data["name"], label_data["description"], label_data["color"]
+        )
+        temp_dir = self.get_scratch_space()
+        with tempfile.NamedTemporaryFile(
```
heanhsok, line 33: Is this `NamedTemporaryFile` needed since we're already using the scratch space?

dev_scripts_helpers/github/test/test_dockerized_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,49 @@
...
+            # Load the label.
+            output_labels = dshgdsgil.Label.load_labels(tmp_file_name)
+            self.assertEqual(len(output_labels), 1)
+            output_label = output_labels[0]
+            # Check.
+            self.assertEqual(output_label.name, input_label.name)
+            self.assertEqual(output_label.description, input_label.description)
+            self.assertEqual(output_label.color, input_label.color)
+        finally:
+            os.remove(tmp_file_name)
```
heanhsok, line 49: No need to remove if we're using the scratch space. It will clean up after itself

dev_scripts_helpers/github/test/test_dockerized_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,49 @@
...
+# #############################################################################
+# TestLabel1
+# #############################################################################
+
+
+class TestLabel1(hunitest.TestCase):
+
+    def test_label_save_and_load(self) -> None:
+        """
+        Test saving and loading a label object.
```
heanhsok, line 25: Same comment as the other test. For unit tests, we usually don't want to test multiple things in the same test case. In this case, if the test fails, without examining the code, it's hard to know if there's an issue with saving or loading the object.

dev_scripts_helpers/github/test/test_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,71 @@
...
+import dev_scripts_helpers.github.sync_gh_issue_labels as dshgsgila
+import helpers.hgit as hgit
+import helpers.hserver as hserver
+import helpers.hunit_test as hunitest
+
+_LOG = logging.getLogger(__name__)
+
+
+# #############################################################################
+# Test_sync_gh_issue_labels1
```
heanhsok, line 16: While testing the read/write operations is good, I think it would be great if we could test the actual logic (which is the most important part of the script).  For example, what happens when
- a label exists in the yaml file but not in the repo
- a label exists in both the file and the repo, but the colors or descriptions are different
- a label exists in the repo but not in the file

heanhsok (changes requested):

dev_scripts_helpers/github/test/test_sync_gh_issue_labels.py
heanhsok: Do you have issue with running Docker inside the container? This fix https://github.com/causify-ai/helpers/pull/583 that @aangelo9 did might help

dev_scripts_helpers/github/test/test_sync_gh_issue_labels.py
sandeepthalapanane: I could change as you suggested, but I cannot test locally due to an issue with the Docker Daemon while building the Dockerized executable. Is there another way I can test this?

dev_scripts_helpers/github/test/test_sync_gh_issue_labels.py
sandeepthalapanane: @heanhsok I tried the fix, but it didn't work. In this [comment](https://github.com/causify-ai/helpers/issues/578#issuecomment-2839860005), you mentioned that only Ubuntu 24.04 LTS is supported, while I am using Ubuntu 20.04.

heanhsok: Since the PR is not yet merged and the image is not yet updated, could you also try to rebuild the container and use the local version? 

This is the commands: https://github.com/causify-ai/helpers/pull/583#discussion_r2067460492

The version that we mentioned is what we have tested and know that it's working, so if there's an error, we can easily reproduce and find the fix quickly. It may work on another distro version as well.

heanhsok: One more thing, I'm thinking, when we test the logic of the script as mentioned in https://github.com/causify-ai/helpers/pull/649#discussion_r2074691009, we probably don't need to have it run in a child container or sibling container. We can just mock the external GH client lib and check the logic. Would that work?

aangelo9: @sandeepthalapanane I recommend upgrading to Ubuntu 24.04 and Docker v28 if possible, since the update has significant internal and workflow changes.

sandeepthalapanane: > One more thing, I'm thinking, when we test the logic of the script as mentioned in [#649 (comment)](https://github.com/causify-ai/helpers/pull/649#discussion_r2074691009), we probably don't need to have it run in a child container or sibling container. We can just mock the external GH client lib and check the logic. Would that work?

This is what I planing to do: mock GitHub in `test_dockerized_sync_gh_issue_labels.py` and check the operations along with the `Label` class since all these utilities are in `dockerized_sync_gh_issue_labels.py`. I would like to keep `test_sync_gh_issue_labels.py` just to verify that the docker image is building properly and to ensure we have a backup file. I want to make sure the test and the utility scripts match.

heanhsok: SG. Feel free to proceed with the plan.

heanhsok (changes requested):
Nice work! Just a couple of small nits.

gpsaggese (commented):
Nicely done!

dev_scripts_helpers/github/test/test_sync_gh_issue_labels.py
```diff
@@ -0,0 +1,81 @@
...
+# #############################################################################
+
+
+class Test_sync_gh_issue_labels1(hunitest.TestCase):
+
+    @pytest.mark.skipif(
+        hserver.is_inside_ci() or hserver.is_dev_csfy(),
+        reason="Disabled because of CmampTask10710",
```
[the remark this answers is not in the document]
sandeepthalapanane: I just upgraded to Ubuntu 24 and will be testing it out. I believe this test is important to ensure that the Docker container functions without any issues. Since we have already tested the functionalities in the dockerized test script, I want to verify that the Docker container operates smoothly as well. This way, we can ensure that everything runs smoothly and ticks all the boxes.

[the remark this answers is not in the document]
heanhsok: SG. It's always good to be thorough

dev_scripts_helpers/github/test/test_sync_gh_issue_labels.py
sandeepthalapanane: @aangelo9 @heanhsok I attempted the fix, built version 1.30 docker image, and modified /etc/sudoers, but I can still not run the dockerized test or perform linting.

When I run the command `i lint --files="` or `i run_fast_tests` I see the password prompt, and when I press `Ctrl + C` it says password required and exits.

```bash
PYTHONPATH=/app:
# Configure env
# set_up_docker_in_docker()
[sudo] password for ubuntu: 
sudo: a password is required
[sudo] password for ubuntu: 
sudo: a password is required
```

aangelo9: Oh, my mistake. Lint uses the `prod` container, tag your locally built `dev` container as prod. Then try linting.
```
docker tag causify/helpers:dev causify/helpers:prod
```

dev_scripts_helpers/github/test/test_sync_gh_issue_labels.py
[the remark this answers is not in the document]
sandeepthalapanane: When I run `test_sync_gh_issue_labels.py` locally, it raises a mount issue because `i run_fast_tests` is already running in a test Docker container. The bind source path does not exist because it is already mounted to the test docker container, which causes the sibling container created by `sync_gh_issue_labels.py` to fail to build.

```bash
RuntimeError: 
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
_system failed: cmd='(docker run --rm --user $(id -u):$(id -g) -e AM_CONTAINER_VERSION -e CSFY_AWS_ACCESS_KEY_ID -e CSFY_AWS_DEFAULT_REGION -e CSFY_AWS_PROFILE -e CSFY_AWS_S3_BUCKET -e CSFY_AWS_SECRET_ACCESS_KEY -e CSFY_AWS_SESSION_TOKEN -e CSFY_CI -e CSFY_ECR_BASE_PATH -e CSFY_ENABLE_DIND -e CSFY_FORCE_TEST_FAIL -e CSFY_GIT_ROOT_PATH -e CSFY_HELPERS_ROOT_PATH -e CSFY_HOST_GIT_ROOT_PATH -e CSFY_HOST_NAME -e CSFY_HOST_OS_NAME -e CSFY_HOST_OS_VERSION -e CSFY_HOST_USER_NAME -e CSFY_REPO_CONFIG_CHECK -e CSFY_REPO_CONFIG_PATH -e CSFY_TELEGRAM_TOKEN -e CSFY_USE_HELPERS_AS_NESTED_MODULE -e OPENAI_API_KEY -e GITHUB_TEST_TOKEN -e PYTHONPATH=/app --workdir /app --mount type=bind,source=/home/sandeep-thalapanane/src/helpers1,target=/app tmp.sync_gh_issue_labels.x86_64.391e0adc /app/dev_scripts_helpers/github/dockerized_sync_gh_issue_labels.py --input_file /app/dev_scripts_helpers/github/test/outcomes/Test_sync_gh_issue_labels1.test1/input/test_gh_issues_labels.yml --owner test-org --repo test-repo --token_env_var GITHUB_TEST_TOKEN --backup --no_interactive) 2>&1'
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
truncated output=
docker: Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /home/sandeep-thalapanane/src/helpers1
```

When I make the dockerized container a child container while testing to resolve the mount issue, the scope of `email_1("dev_scripts_helpers.github.dockerized_sync_gh_issue_labels.github.Github")` becomes extinct. As a result, the mocking does not occur, causing the test to fail.

Its like a loop, solving one error leads to the other error.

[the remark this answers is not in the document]
sandeepthalapanane: I tested the mocking logic outside the Docker container, and it works. Once the CmampTask10710 is fixed, I will verify if the docker functions correctly.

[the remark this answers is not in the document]
heanhsok: SG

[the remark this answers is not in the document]
heanhsok: Feel free to leave a TODO in the comment.

sandeepthalapanane:
@heanhsok I've completed all the changes. Please review the changes.

heanhsok (approved):

PR #651, opened 2025-05-02

sandeepthalapanane:
Addressing #615 

Added unit tests for `docker_build_prod_image` (both single arch and multi-arch)


helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -182,3 +179,250 @@ def test_docker_build_multi_arch(self) -> None:
...
             dedent=True,
         )
+
+
+# #############################################################################
+# TestDockerBuildProdImage1
+# #############################################################################
+
+
+class TestDockerBuildProdImage1(hunitest.TestCase):
```
gpsaggese, line 189: let's call it Test_docker_build_prod_image1 so it's easier to grep.
You can rename the others too.

We are ok to override PEP8 rules here.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -182,3 +179,250 @@ def test_docker_build_multi_arch(self) -> None:
...
+
+    def setUp(self) -> None:
+        """
+        Set up test environment and initialize all necessary mocks.
+        """
+        super().setUp()
+        # Mock system calls.
+        self.system_patcher = umock.patch("helpers.hsystem.system")
```
gpsaggese, line 197: Can we factor out some of the same code across tests?

It's not only about DRY but also because there are common idioms

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -182,3 +179,250 @@ def test_docker_build_multi_arch(self) -> None:
...
+        """
+        Test building a prod Docker image with single architecture.
+
+        This test verifies that the correct sequence of commands is
+        generated for building a prod Docker image.
+        """
+        # Prepare inputs.
+        mock_ctx = httestlib._build_mock_context_returning_ok()
+        test_version = "1.0.0"
+        test_base_image = "test-registry.com/test-image"
```
gpsaggese, line 239: Can this be a function?

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -182,3 +179,250 @@ def test_docker_build_multi_arch(self) -> None:
...
+        # Call tested function.
+        hltadore.docker_build_prod_image(
+            mock_ctx,
+            test_version,
+            base_image=test_base_image,
+            cache=False,
+        )
+        actual_cmds = _extract_commands_from_call(self.mock_run.call_args_list)
```
gpsaggese, line 247: You can create a function with the common idiom for checking the output

def check_docker_command_output(self_: Any, exp: str, ...)

gpsaggese (commented):
Going in the right direction, but the amount of copy-paste gives the reader headache.

There are some chunks of code that have a clear meaning and should be factored out
Let's factor them out.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -182,3 +179,250 @@ def test_docker_build_multi_arch(self) -> None:
...
+        # Check output.
+        self.assert_equal(
+            actual_cmds,
+            exp,
+            purify_text=True,
+            purify_expected_text=True,
+            fuzzy_match=True,
+            remove_lead_trail_empty_lines=True,
+            dedent=True,
+        )
```
dremdem, line 428: I'd make it just: 
```
        self.assert_equal(
            actual_cmds,
            exp,
            fuzzy_match=True,
        )
```

Reason: no need to purify anything since we are using the test examples.

dremdem: Same everywhere.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -182,3 +179,250 @@ def test_docker_build_multi_arch(self) -> None:
...
+
+    def tearDown(self) -> None:
+        """
+        Clean up test environment by stopping all mocks after each test case.
+        """
+        self.system_patcher.stop()
+        self.run_patcher.stop()
+        self.version_patcher.stop()
+        self.docker_login_patcher.stop()
+        self.env_patcher.stop()
```
dremdem, line 226: Perhaps, it will be nice to put them to a list, and then stop with a loop.
Your call.

dremdem (commented):
LG, just couple nits.

helpers/test/test_lib_tasks_docker_release.py
sandeepthalapanane: I factored out common code into a helper test class.

helpers/test/test_lib_tasks_docker_release.py
sandeepthalapanane: When I removed `purify_text`, a few tests failed.

helpers/test/test_lib_tasks_docker_release.py
sandeepthalapanane: Changed it to a list.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -35,12 +36,30 @@ def _extract_commands_from_call(calls: List[umock._Call]) -> List[str]:
     return call_list
 
 
+def get_build_inputs() -> Tuple[invoke.MockContext, str, str, str]:
```
dremdem, line 39: -> `_ get_build_inputs()`

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -63,19 +82,57 @@ def setUp(self) -> None:
...
 
     def tearDown(self) -> None:
         """
         Clean up test environment by stopping all mocks after each test case.
         """
-        self.system_patcher.stop()
-        self.run_patcher.stop()
-        self.version_patcher.stop()
-        self.docker_login_patcher.stop()
+        patchers = [
```
dremdem, line 95: I'd move it to `self.patchers` in the `setUp()`.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -35,12 +36,30 @@ def _extract_commands_from_call(calls: List[umock._Call]) -> List[str]:
...
+    Create common test inputs for docker build flow tests.
+
+    :return: tuple containing mock context, test version, test base
+        image and test multi-arch
+    """
+    mock_ctx = httestlib._build_mock_context_returning_ok()
+    test_version = "1.0.0"
+    test_base_image = "test-registry.com/test-image"
+    test_multi_arch = "linux/amd64,linux/arm64"
+    return mock_ctx, test_version, test_base_image, test_multi_arch
```
dremdem, line 50: I'd make them as `dict`: 
```
inputs[
   "mock_ctx": ...,
   "version": ...,
   ...
]
```

dremdem: btw, we can move them to the `setUp()` and use everywhere as `self.mock_ctx, ..., ...`.
WDYT?

dremdem (commented):
Good progress.
Some comments.

helpers/test/test_lib_tasks_docker_release.py
sandeepthalapanane: These tests are also failing. Since I created the same function for assertion check for both `local` and `prod`, we should keep `purify_txt` to ensure it works across tests.

helpers/test/test_lib_tasks_docker_release.py
sandeepthalapanane: I think we could move it to `setUp`, which would also be useful for testing the tagging, pushing, and roll_back functions.

dremdem (approved):
LGTM

sandeepthalapanane:
@gpsaggeseI I have completed creating tests for almost all the functions in `lib_tasks_docker_release.py`. Would you like to push these tests in this PR itself or create a separate PR after merging this one?

gpsaggese (commented):
Good work

gpsaggese:
> @gpsaggese I have completed creating tests for almost all the functions in `lib_tasks_docker_release.py`. Would you like to push these tests in this PR itself or create a separate PR after merging this one?

Sorry I didn't see this. When you are blocked just send me a ping email with the PR to review, etc.

Yes, now let's test all the ones we can test. We don't want to overtest, so if you want you can create a test plan for us to review.

Issue #622, opened 2025-04-27
gitleaks doesn't run in submodules


gpsaggese:
When running gitleaks in githooks, I get

```
##### check_gitleaks ######

    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

2:44PM ERR [git] fatal: not a git repository: /app/../.git/modules/helpers_root
2:44PM ERR failed to scan Git repository error="stderr is not empty"
2:44PM WRN scanned ~0 bytes (0)
2:44PM WRN partial scan completed in 46.2ms
2:44PM WRN no leaks found in partial scan
```

The problem is that submodules are not real git repo and Gitleak gets unhappy.

The solution we usually use is to find the root of the submodule to mount the outermost Git module and then cd in the submodule to do the work.

For now I'll disable the gitleaks step when in a submodule

heanhsok:
I've changed to the outermost git root dir in this PR [654](https://github.com/causify-ai/helpers/pull/654)

gpsaggese:
After merging the PR now I'm getting a different error.

I'm committing from `helpers` as sub-repo and it looks like 

```
##### check_gitleaks ######

    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

10:17PM FTL unable to load gitleaks config, err: open /app/.github/gitleaks-rules.toml: no such file or directory

'check_gitleaks' failed
Warning: Permanently added 'github.com' (ED25519) to the list of known hosts.
Everything up-to-date
```

The file is in helpers but not in the including repo.

```
saggese@gpmac.local ==> RC: 1 <== venv:(client_venv.helpers) branch:'CmampTask11862_Allow_dind_unit_tests_to_run_on_server_and_CI' ~/src/notes1/helpers_root
> ls .github/gitleaks-rules.toml
.github/gitleaks-rules.toml

saggese@gpmac.local venv:(client_venv.helpers) branch:'CmampTask11862_Allow_dind_unit_tests_to_run_on_server_and_CI' ~/src/notes1/helpers_root
> ls ../.github/gitleaks-rules.toml
ls: ../.github/gitleaks-rules.toml: No such file or directory
```

1) So we need to have a `.github/gitleaks-rules.toml` in every repo (which makes sense). Correct?
2) Can we ensure this info is in the "template" / instructions?
3) Should we use links in the future to keep these files in sync?

heanhsok:
[...]

PR #654, opened 2025-05-04
HelpersTask622_gitleaks_doesnt_run_in_submodules

heanhsok:
Task https://github.com/causify-ai/helpers/issues/622


gpsaggese (commented):
Looks good.

Do we have a bounty for testing this, right?
How can we test this part?
My only idea is to have a unit test (outside pytest) for all the repos to make sure everything is properly configured.

PR #664, opened 2025-05-05
HelpersTask622_gitleaks_doesnt_run_in_submodules_2

heanhsok:
Task #622

- Move gitleaks rule from `.github` to `dev_scripts_helpers/git/gitleaks` so it can be pointed to from both the GH actions and the git hooks without creating symlinks


dev_scripts_helpers/git/git_hooks/utils.py
```diff
@@ -468,29 +469,20 @@ def check_python_compile(
...
-
 def check_gitleaks(abort_on_error: bool = True) -> None:
     """
     Check that the code does not contain any leaked secrets.
     """
     func_name = _report()
-    git_root_dir = get_git_root_dir()
+    # Find relative path from the git root to the helpers root.
+    git_root_dir = hgit.find_git_root()
+    helpers_root_dir = hgit.find_helpers_root()
```
heanhsok, line 479: - I realize we'll be copying a lot of code from hgit for these two functions
- Since Git hooks are now installed and run after our thin client is activated, should we consider using the helpers module?
- This script is already somewhat coupled to the helpers, especially with the introduction of the gitleaks config path. 
- To avoid this coupling, we can also just keep things the way they are and make sure that each repo has this `.github/gitleaks-rules.toml` file

WDYT? @gpsaggese

gpsaggese: Agree. Let's just import helpers and embrace it that a piece of the infra checking on the code depends on the code itself.

In an ideal world the git hooks should be a prod container (like the linter), but it is too paranoid

Issue #393, opened 2025-03-19
Review systems to automate code review


gpsaggese:
We want to find a way to do the first line of defense for our PRs

Some ideas
- https://docs.github.com/en/copilot/using-github-copilot/code-review/using-copilot-code-review
- graphite.dev
- https://techcrunch.com/2019/08/06/deepcode-gets-4m-to-feed-its-ai-powered-code-review-tool/
- https://techcrunch.com/2024/08/15/coderabbit-raises-16m-to-bring-ai-to-code-reviews/

Shaunak01:
Some info: https://docs.google.com/document/d/1cs50rwqxhRzQLt41brcPoQs5xL0QYOEFDcKHJB0NUo8/edit?tab=t.0#heading=h.rv0yjq809l2h

gpsaggese:
Okay, let's start with Copilot and then try graphite. What's the next step? @Shayawnn, do you have the permissions to enable it or is it already on?

gpsaggese:
@Shaunak01 can you change the name of the gdoc as HelpersTaskXYZ_... same as a PR. This is how we keep the gdoc in sync with the tasks

gpsaggese:
We want a "bot" to add comments to the PR related to:
1) stylistic stuff (based on our code style)
2) look for mistakes
...
to automate the "first" PR

You can also set it up on your side and we do a demo to decide which one is best and why

gpsaggese:
This is in progress right? If so, let's mark it

Shaunak01:
@gpsaggese Codecov has sent request to cmamp to get access. Can you check

gpsaggese:
Accepted. We can start from `//helpers` since the regressions are faster

Shaunak01:
@gpsaggese 

https://github.com/github-copilot/purchase

When you activate copilot pro for the organization we should be able to use Copilot for automated code reviews (we need to add as a reviewer if not set as default) and other features as well. No setup required from git actions or code base.

gpsaggese:
@gpsaggese 
- [ ] Read gdoc
- [ ] Enable copilot

gpsaggese:
Good work here.

Next steps are:
1) Convert the doc into a markdown with our script
    - I would remove the tables since they are difficult to render / edit in md and put them in bullet lists
2) Instead of the pic, for pricing let's summarize
    - Link to the page
    - Cost per user, 10 users, 20 users
[...]

PR #710, opened 2025-05-13
Helpers task393 review systems to automate code review 2

sonniki:
#393 

Adding an "LLM-friendly" doc with our conventions and guidelines both for the PR workflows and the checked in code.
This doc lists them in a much more concise format than our regular docs, and we can try to use it as input to the automated code review tools.

### DIFF
--- a/docs/code_guidelines/all.automated_review_guidelines.reference.md
+++ b/docs/code_guidelines/all.automated_review_guidelines.reference.md
@@ -1,0 +1,489 @@
+# Guidelines for automated PR reviews
+
+<!-- toc -->
+
+- [PR workflows](#pr-workflows)
+  * [Working in a branch](#working-in-a-branch)
+  * [Filing a PR](#filing-a-pr)
+  * [Before requesting review](#before-requesting-review)
+  * [During the review cycle](#during-the-review-cycle)
+- [Code style](#code-style)
+  * [Python code](#python-code)
+    + [Naming](#naming)
+    + [Docstrings](#docstrings)
+    + [Comments](#comments)
+    + [Code design](#code-design)
+    + [Imports](#imports)
+    + [Type annotations](#type-annotations)
+    + [Functions](#functions)
+    + [Scripts](#scripts)
+    + [Logging](#logging)
+    + [Unit tests](#unit-tests)
+    + [Misc](#misc)
+  * [Notebooks](#notebooks)
+    + [General](#general)
+    + [Jupytext](#jupytext)
+    + [Plotting](#plotting)
+  * [Markdowns](#markdowns)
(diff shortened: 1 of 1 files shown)


docs/code_guidelines/all.automated_review_guidelines.reference.md
```diff
@@ -0,0 +1,484 @@
...
+<!-- tocstop -->
+
+This document outlines the rules that all PRs and checked-in code must follow.
+It can serve as a guideline for automated PR reviews.
+
+## PR workflows
+
+### Working in a branch
+
+- The branch should be named following the format
```
gpsaggese, line 40: The LLM can't easily check (unless it's agentic and can run git, aider can).
We are going to put it in a Git hook to enforce that. There is an issue for that being worked on. 

How about we create a PR checklist, that humans run on their side before the PR?
You can add here.
./helpers_root/docs/code_guidelines/all.submit_code_for_review.how_to_guide.md

Then we force people to post it for every review, and check the boxes, and we use it as a blueprint for automation.

docs/code_guidelines/all.automated_review_guidelines.reference.md
```diff
@@ -0,0 +1,484 @@
...
+- All the checks performed by GitHub Actions must pass
+- The branch must be up to date with the master branch
+- There should be no conflicts with the master branch
+- There should be no files checked in by mistake (such as tmp and log files)
+- All checked in files should be checked and formatted by Linter
+- No files larger than 500 KB should be checked in
+- Screenshots should not be used in PR posts to describe the situation or report
+  an error — copy-and-paste instead
+
+### During the review cycle
```
gpsaggese, line 73: All this should go in the PR checklist

docs/code_guidelines/all.automated_review_guidelines.reference.md
```diff
@@ -0,0 +1,484 @@
...
+
+- Label `PR_for_reviewers` should be present when a review is requested
+- Fixes addressing a review comment should be applied everywhere, not just where
+  the reviewer pointed out the issue
+- After addressing a review comment, the corresponding conversation should be
+  marked as `resolved`
+- After all review comments are resolved, `re-request review` button should be
+  used to request another round of review
+
+## Code style
```
gpsaggese, line 83: This is all very good. Let's keep it here and start experimenting with llm_transform targets.

We can have an executable `ai_reviewer.py` which loads this file and looks for violation (and then can try to fix, but this is difficult due to the large number of false positivies)

gpsaggese (commented):
Good work! Let's merge and we can keep working. I'll file some follow up issues

PR #733, opened 2025-05-17

sandeepthalapanane:
Addressing #615 

Added tests for all the remaining functions.

The function `docker_release_all()` simply calls two other functions that have already been tested. Rerunning tests for this function without skipping (since there is no argument to skip) causes issues, so I decided to ignore testing it. This function has already been mentioned as unnecessary in the TODO list, so I'm skipping the test for this function.

Similarly, `docker_create_candidate_image()` also relies on two thoroughly tested functions and one extra command. I completely mocked these functions and verified only the final single command.


helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -796,3 +800,383 @@ def test_aws_ecr1(self) -> None:
...
+# Test_docker_create_candidate_image1
+# #############################################################################
+
+
+class Test_docker_create_candidate_image1(_DockerFlowTestHelper):
+    """
+    Test creating a candidate Docker image.
+    """
+
+    def set_up_test2(self) -> None:
```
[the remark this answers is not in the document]
dremdem: Same everywhere.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -796,3 +800,383 @@ def test_aws_ecr1(self) -> None:
...
+    @pytest.fixture(autouse=True, scope="class")
+    def aws_credentials(self) -> Generator[None, None, None]:
+        """
+        Mocked AWS credentials for moto.
+        """
+        os.environ["MOCK_AWS_ACCESS_KEY_ID"] = "testing"
+        os.environ["MOCK_AWS_SECRET_ACCESS_KEY"] = "testing"
+        os.environ["MOCK_AWS_SECURITY_TOKEN"] = "testing"
+        os.environ["MOCK_AWS_SESSION_TOKEN"] = "testing"
+        os.environ["MOCK_AWS_DEFAULT_REGION"] = "us-east-1"
```
dremdem, line 1003: Can you take a look at `helpers/test/test_haws.py`.
Not sure, but perhaps we can re-use it here.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -796,3 +800,383 @@ def test_aws_ecr1(self) -> None:
...
+
+        This test checks:
+        - Task definition update workflow
+        - Preprod to prod image conversion.
+        - DAG file synchronization
+        - Image tagging and pushing
+        """
+        # Mock ECS client and task definition.
+        region = "us-east-1"
+        mock_client = boto3.client("ecs", region_name=region)
```
dremdem, line 1081: Need to specify which client.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -796,3 +800,383 @@ def test_aws_ecr1(self) -> None:
...
+        self.patchers.append(self.ecs_client_patcher)
+        # Call tested function.
+        hltadore.docker_update_prod_task_definition(
+            self.mock_ctx,
+            version=self.test_version,
+            preprod_tag="test_hash",
+            airflow_dags_s3_path="s3://test-bucket/dags/",
+            task_definition="test_task",
+        )
+        exp = r"""
```
dremdem, line 1111: It would be nice to check the arguments passed to `haws.update_task_definition()` here.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -796,3 +800,383 @@ def test_aws_ecr1(self) -> None:
...
+        )
+        mock_get_ecs_client.return_value = mock_client
+        # Add mock client to patchers for cleanup
+        self.ecs_client_patcher = umock.patch(
+            "boto3.client", return_value=mock_client
+        )
+        self.mock_ecs_client = self.ecs_client_patcher.start()
+        self.patchers.append(self.ecs_client_patcher)
+        # Mock S3 bucket operations to simulate a failure.
+        self.mock_s3.return_value.put.side_effect = Exception("S3 upload failed")
```
dremdem, line 1161: Can you also check the `S3Mock_TestCase` in `helpers/hmoto.py`? Not sure, but it might be the case.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -796,3 +800,383 @@ def test_aws_ecr1(self) -> None:
...
+        docker tag test.ecr.path/test-image:test_hash test.ecr.path/test-image:prod
+        docker rmi test.ecr.path/test-image:test_hash
+        docker push test.ecr.path/test-image:prod-1.0.0
+        docker push test.ecr.path/test-image:prod
+        """
+        self._check_docker_command_output(exp, self.mock_run.call_args_list)
+
+    @moto.mock_aws
+    @umock.patch("helpers.haws.get_ecs_client")
+    def test_promotion_to_prod_exception_handling(
```
dremdem, line 1123: Nice idea!

dremdem (commented):
Solid work. 
Some suggestions.

helpers/test/test_lib_tasks_docker_release.py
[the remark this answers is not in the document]
sandeepthalapanane: I have used `set_up_test2()` as per guidelines in the https://github.com/causify-ai/helpers/blob/master/docs/tools/unit_test/all.write_unit_tests.how_to_guide.md#L757.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -796,3 +800,383 @@ def test_aws_ecr1(self) -> None:
...
+        Set up test environment with additional mocks specific to this test
+        class.
+        """
+        self.set_up_test()
+        # Mock git hash.
+        self.git_hash_patcher = umock.patch(
+            "helpers.hgit.get_head_hash", return_value="test_hash"
+        )
+        self.mock_git_hash = self.git_hash_patcher.start()
+        self.patchers.append(self.git_hash_patcher)
```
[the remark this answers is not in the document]
sandeepthalapanane: Sure, we could convert it to dict to make it more readable.

helpers/test/test_lib_tasks_docker_release.py
sandeepthalapanane: I tried reusing, which caused `TestGenerateAwsFiles.test1` to fail in `helpers/test/test_haws.py`. That’s why I added a cleanup of the os environment variable to ensure that tests are isolated from one another.

helpers/test/test_lib_tasks_docker_release.py
sandeepthalapanane: The test covers a full task promotion flow, which involves both ECS (task definition updates) and S3 (DAG uploads), so both services are being mocked using moto and unittest.mock as needed. S3Mock_TestCase seems tailored for isolated S3 testing, while the test handles orchestration logic involving ECS clients and error handling across services.

dremdem (approved):
LGTM. 
Feel free to address one nit and request a review from Samarth.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -796,3 +800,393 @@ def test_aws_ecr1(self) -> None:
...
+            "helpers.hgit.get_head_hash",
+            return_value="4759b3685f903e6c669096e960b248ec31c63b69",
+        )
+        self.mock_git_hash = self.git_hash_patcher.start()
+        self.patchers["git_hash"] = self.git_hash_patcher
+        # Mock workspace size check.
+        self.workspace_check_patcher = umock.patch(
+            "helpers.lib_tasks_docker_release._check_workspace_dir_sizes"
+        )
+        self.mock_workspace_check = self.workspace_check_patcher.start()
```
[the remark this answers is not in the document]
sandeepthalapanane: I ran into an issue because of not mocking `self.mock_workspace_check`. I think it's a good practice to check only the Docker command rather than verifying large files, and it also runs a system command, which takes time.

Issue #721, opened 2025-05-15
Little fix in purify for /app


gpsaggese:
Good debugging. Then it's a bug inside the purification. I'll file a bug and you can do a quick fix

_Originally posted by @gpsaggese in https://github.com/causify-ai/helpers/pull/707#discussion_r2091504510_

PR #734, opened 2025-05-17
HelpersTask721_Little_fix_in_purify_for_app

sandeepthalapanane:
Fixed issue #721 by updating `purify_app_references()` to correctly process `/app`. 

I adjusted the sequence of function calls in `purify_txt_from_client()` to ensure that `/app` is passed to `purify_app_references()` first, which replaces it with `""`. If we pass `/app` to `purify_from_environment()`, it first converts `/app` into `$GIT_ROOT`.


helpers/hunit_test.py
```diff
@@ -617,8 +617,8 @@ def purify_txt_from_client(txt: str) -> str:
     """
     Remove from a string all the information of a specific run.
     """
-    txt = purify_from_environment(txt)
     txt = purify_app_references(txt)
+    txt = purify_from_environment(txt)
```
gpsaggese, line 621: If there is an order we should explain why and how.

It's best to make all the calls idempotent to avoid these problems

helpers/hunit_test.py
```diff
@@ -455,7 +455,7 @@ def purify_app_references(txt: str) -> str:
     """
     Remove references to `/app`.
     """
-    txt = re.sub(r"/app/", "", txt, flags=re.MULTILINE)
+    txt = re.sub(r"/app$", "", txt, flags=re.MULTILINE)
```
gpsaggese, line 458: Uhm ... the original code was not great, but now it's even more overfit to your case.

gpsaggese (commented):
Your fix is valid, but that code was kind of horrible already. Let's chat in the next sync on how to clean it up a bit.

helpers/hunit_test.py
sandeepthalapanane: I attempted to make the process idempotent; however, there are many redundancies between `purify_from_environment` and `purify_app_references`, which complicates this task. Would you like me to change the code in both functions to eliminate these redundancies in this pull request itself, or should I create a new pull request for those changes?

PR #738, opened 2025-05-19
Helpers task727 generalise price counting interface in hopenai.py

srinivassaitangudu:


gpsaggese (commented):
Code is fine. It would be good to test it as well. Maybe we can re-use the caching approach adding more metadata

Issue #749, opened 2025-05-21
Update `parallel_execute` to have flags for log files


Shaunak01:
- Currently the func `parallel_execute` in `hjoblib.py` generates the log file and a lot of info for status by default 
- We want some flags so that we dont hang with logs

FYI @gpsaggese @sonniki

Shaunak01:
PR: https://github.com/causify-ai/helpers/pull/751

sonniki:
Done

PR #751, opened 2025-05-21
HelpersTask749_Update_parallel_execute_to_have_flags_for_log_files

Shaunak01:
issue: https://github.com/causify-ai/helpers/issues/749

### DIFF
--- a/helpers/hjoblib.py
+++ b/helpers/hjoblib.py
@@ -552,7 +552,6 @@ def _parallel_execute_decorator(
         txt.append(f"func_res=\n{hprint.indent(str(res))}")
     else:
         txt.append("func_res=<omitted>")
-
     txt.append(f"elapsed_time_in_secs={elapsed_time}")
     txt.append(f"start_ts={start_ts}")
     txt.append(f"end_ts={end_ts}")
--- a/helpers/hjoblib.py
+++ b/helpers/hjoblib.py
@@ -469,7 +469,7 @@ def _parallel_execute_decorator(
     func_name: str,
     processify_func: bool,
     task: Task,
-    enable_logging: bool,
+    enable_file_logging: bool,
     verbose_log: bool,
 ) -> Any:
     """
@@ -483,6 +483,8 @@ def _parallel_execute_decorator(
             - if `abort_on_error=False` the exception is not propagated, but the
               return value is the string representation of the exception
     :param processify_func: switch to enable wrapping a function into a process
+    :param enable_file_logging: see same parameter in `parallel_execute()`
+    :param verbose_log: see same parameter in `parallel_execute()`
     :return: the return value of the workload function or the exception string
     """
     # Validate very carefully all the parameters.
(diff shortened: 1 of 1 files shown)


helpers/hjoblib.py
```diff
@@ -469,6 +469,8 @@ def _parallel_execute_decorator(
     func_name: str,
     processify_func: bool,
     task: Task,
+    enable_logging: bool,
```
gpsaggese, line 472: let's call it enable_file_logging since logging can also be on screen.

helpers/hjoblib.py
```diff
@@ -469,6 +469,8 @@ def _parallel_execute_decorator(
...
+    verbose_log: bool,
```
gpsaggese, line 473: Let's update the docstring with this. Since this private you can point to the doc of the public one

```
:param enable_file_logging: see same parameter in ...
...
```

helpers/hjoblib.py
```diff
@@ -596,17 +605,18 @@ def parallel_execute(
...
-    :param backend: specify the backend type (e.g., joblib `loky` or
-        `asyncio_process_executor`)
-
-    :return: list with the results from executing `func` or the exception of the
-        failing function
-        - NOTE: if `abort_on_error=True` and one task fails, `joblib` doesn't return
-          the output of the already executed tasks. In this case, the best we can do
-          is to return the exception of the failing task
+    :param backend: specify the backend type (e.g., joblib `loky` or `asyncio_process_executor`)
+    :param enable_logging: if False, skip writing any log file (default: True)
```
gpsaggese, line 609: DRY. the default is already in the interface

helpers/hjoblib.py
```diff
@@ -596,17 +605,18 @@ def parallel_execute(
...
+    :param verbose_log: if True, write detailed task results (e.g., embeddings) to the log file
+        - If False, large outputs will be omitted from the log to reduce file size (default: False)
+    :return: list with the results from executing `func` or the exception of the failing function
+    Note:
```
gpsaggese, line 613: Let's move this note to the middle of the docstring to make it more visible

gpsaggese (commented):
Minor things. Good change

helpers/hjoblib.py
```diff
@@ -596,17 +605,18 @@ def parallel_execute(
...
-        `asyncio_process_executor`)
-
-    :return: list with the results from executing `func` or the exception of the
-        failing function
-        - NOTE: if `abort_on_error=True` and one task fails, `joblib` doesn't return
-          the output of the already executed tasks. In this case, the best we can do
-          is to return the exception of the failing task
+    :param backend: specify the backend type (e.g., joblib `loky` or `asyncio_process_executor`)
+    :param enable_logging: if False, skip writing any log file (default: True)
+    :param verbose_log: if True, write detailed task results (e.g., embeddings) to the log file
```
sonniki, line 610: I would not specify "(e.g., embeddings)" here since it might only be confusing to the user who is unfamiliar with our context

sonniki (commented):
Nothing else besides GP's points

sonniki (approved):

Issue #758, opened 2025-05-23
test_aws_ecr1 fails when helpers is a submodule


sandeepthalapanane:
As mentioned in #756, the `Test_docker_create_candidate_image1.test_aws_ecr1` fails when the helpers repo is used as a submodule.

FYI @dremdem @sonniki

sandeepthalapanane:
Merged into master.

Issue #762, opened 2025-05-24
Clean up the purification layer


gpsaggese:
It should be a different PR (and even a different Issue). I would start with adding unit tests to make sure things are sane, and then improve it.

Touching the unit test framework is dangerous business and we need to do it carefully.

But that part is like ~10 years old (no jokes) and quite messy. I appreciate the courage to jump into that

_Originally posted by @gpsaggese in https://github.com/causify-ai/helpers/pull/734#discussion_r2105966418_

gpsaggese:
Good analysis.

1) I don't like the amp / app mix up, it should be separated (at least in terms of code)
2) I am ok with having a precise order of applying the transforms
3) Each replacement should be done as an atom (e.g., the dir `app` should be recognized and replaced wherever it is, as long as it's not inside other strings, e.g., `application`)
4) We should document these design principles in a short doc (and we'll add it to the doc for https://github.com/causify-ai/helpers/issues/779)

sandeepthalapanane:
Merged to master https://github.com/causify-ai/helpers/pull/786

PR #765, opened 2025-05-26
HelpersTask758_test_aws_ecr1_fails_when_helpers_is_a_submodule

sandeepthalapanane:
Fixed the #758 with a submodule check and changed the expected command.


helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -968,9 +968,17 @@ def test_aws_ecr1(self) -> None:
             task_definition="test_task",
             user_tag="test_user",
         )
-        exp = r"""
-        invoke docker_cmd -c "amp/datapull/aws/aws_update_task_definition.py -t test_task -i test_user-4759b3685f903e6c669096e960b248ec31c63b69 -r eu-north-1"
-        """
+        # Check if amp is a submodule.
+        path, _ = hgit.get_path_from_supermodule()
```
dremdem, line 972: It's better to use `hgit.is_in_amp_as_submodule()` for our check.

The full story:
- We have a script `datapull/aws/aws_update_task_definition.py` located in the `cmamp` repo.
- We have another repo that uses `cmamp` (also known as `amp`) as a submodule.
- The `hltadore.docker_create_candidate_image()` is trying to find it in the following paths:
   - `datapull/aws/aws_update_task_definition.py`
   - `amp/datapull/aws/aws_update_task_definition.py`

So, we can either repeat the logic here:
https://github.com/causify-ai/helpers/blob/0387ed7fa93653b9ca98396fb3580a7fa7a19ba3/helpers/lib_tasks_docker_release.py#L1272-L1275

Or use `hgit.is_in_amp_as_submodule()`, which is clearer.

dremdem (commented):
Nice try, but let's go another way.

helpers/test/test_lib_tasks_docker_release.py
sandeepthalapanane: I switched to using `is_in_amp_as_submodule()`. I am also working on changing the purify text functionality in #762 to ensure it gets purified correctly. The `amp` is not removed from the text even though we specify `purify_expected_txt = True` because the purify function only identifies `amp` when it appears at the beginning of the line. I plan to modify it so that it can detect and remove `amp` even when it is located elsewhere. This change should help us avoid these issues in the future.

helpers/test/test_lib_tasks_docker_release.py
```diff
@@ -968,9 +968,17 @@ def test_aws_ecr1(self) -> None:
...
+        is_in_amp_as_submodule = hgit.is_in_amp_as_submodule()
+        # Set the expected docker command based on whether amp is a submodule.
+        if is_in_amp_as_submodule:
+            exp = r"""
+            invoke docker_cmd -c "datapull/aws/aws_update_task_definition.py -t test_task -i test_user-4759b3685f903e6c669096e960b248ec31c63b69 -r eu-north-1"
+            """
+        else:
+            exp = r"""
+            invoke docker_cmd -c "amp/datapull/aws/aws_update_task_definition.py -t test_task -i test_user-4759b3685f903e6c669096e960b248ec31c63b69 -r eu-north-1"
+            """
```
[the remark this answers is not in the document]
sandeepthalapanane: When I run tests locally, the actual command generated includes the `amp/` prefix, even though `amp` is not a submodule. In the implementation of `helpers/lib_tasks_docker_release.py`, the `amp/` prefix is added when the direct path does not exist. Since we are running the tests in a container and the file is not present, it adds `/amp` prefix, but what we want is probably `/app`. 

To achieve consistency, especially when we are mocking `hdbg.dassert_file_exists`, we may need to modify the code accordingly, as this will ensure the behavior remains the same even if `helpers` is treated as a submodule.

```
if not os.path.exists(exec_name):
        exec_name = f"amp/{exec_name}"
```

to this 
``` 
if not hdbg.dassert_file_exists(exec_name):
        exec_name = f"amp/{exec_name}"
```

sandeepthalapanane:
@dremdem In the latest master branch update https://github.com/causify-ai/helpers/commit/bbba40974dec4b06dc9db35e2ac95842819499fe, I noticed the removal of the code that adds the /amp prefix in `lib_tasks_docker_release.py`. The test file is also updated, so we could close this PR and the issue as it is not needed anymore.

dremdem (approved):
LGTM

Issue #717, opened 2025-05-14
Create AI reviewer script


gpsaggese:
We want to create a system similar to `llm_transform`, called `ai_reviewer.py` that

1) Read the file from https://github.com/causify-ai/helpers/pull/710#discussion_r2089201962

2) Run the checks from the file and create a file similar to the linter with the violations to the our rules from the guidelines

3) Then we can use `llm_apply.py` to fix the problems automatically but this requires human supervision. The flow I use is something like:
    - Use `ai_reviewer.py` to find the problems as cfile
    - Jump around accepting / rejecting the suggestions
    - Use `llm_apply.py` to utilize LLMs to address the issues (we can apply changes in multiple commits, e.g., by type of violation). It seems that giving all the rules at once, confused the LLMs

Most of the code (if not everything) is already in `llm_transform.py` and `llm_apply.py`, we just have to stitch things together.

Let me know if you want PP to build the skeleton. Then we can start outsourcing testing and adding more stuff to the good collaborators

gpsaggese:
Some notes from this AM discussion
- The goal is to create a script ai_reviewer.py
   - Automate the usual complaints
   - People run the script, fix the code by hand (or use llm_apply to fix them, but then they won't learn)
   - Submit code that doesn't upset the reviewers too much

gpsaggese:
Merged the script from @sonniki 

Next steps are
- Merge my code on top of it
   - inject_todo.py
   - Split the md files into chunks for different "targets" (e.g., linter, llm, correctness)

sonniki:
Various AI reviewer and code editing scripts are collected in https://github.com/causify-ai/helpers/tree/master/dev_scripts_helpers/llms
Also, I suppose, CC has become a big part of achieving the original goals.
Don't think there is anything in particular that needs to be done here, so, closing.

PR #766, opened 2025-05-27
Helpers task717 create ai reviewer script

sonniki:
#717 

Adding a script with an LLM reviewer that checks if the code in a given file violates any guidelines listed in a Markdown. It outputs violations in a Linter style, saves them into a log file and prints in the terminal. 
The input arguments follow the Linter style as well (with options like `--files`, `--modified`, etc). 
The guidelines in the Markdown are updated to follow "one guideline per bullet point" format in order to make it easier to pass to the LLM.

Limitation: it's extremely slow, since it makes a separate LLM call for every guideline (e.g. 50+ guidelines for Python files). Probably need to change this logic somehow.

Also -- it hallucinates A LOT.


gpsaggese (commented):
Looks good

PR #768, opened 2025-05-27
CMTask12257: Remove git root assertion

mongolianjesus:
[#12257](https://github.com/causify-ai/cmamp/issues/12257)

Removed the git root assertion from thin client script.


sonniki:
Merging to unblock development. 
@gpsaggese if anything is not right, follow-ups can be filed

Issue #772, opened 2025-05-28
Use return 1 to stop the execution in shell script


heanhsok:
Follow up from https://github.com/causify-ai/helpers/issues/726#issuecomment-2884901652
> 2. How come the assertion doesn't stop the execution?
> 
> I think maybe it’s because we're using `kill -INT $$` to stop the script. From what I have read, if the parent shell traps or ignores SIGINT, that signal won’t terminate the script. Using `exit 1` or `return 1` might be more reliable?
> 
> https://github.com/causify-ai/helpers/blob/4868c54a4abf5a45ebef98dc7574422408e6bcf9/dev_scripts_helpers/thin_client/thin_client_utils.sh#L56-L63 

 _Originally posted by @heanhsok in [#726](https://github.com/causify-ai/helpers/issues/726#issuecomment-2884901652)_

FYI @gpsaggese

PR #773, opened 2025-05-28
HelpersTask772_Use_return__1_to_stop_the_execution_in_shell_script

heanhsok:
Task #772


dev_scripts_helpers/thin_client/thin_client_utils.sh
```diff
@@ -31,7 +31,7 @@ dassert_is_sourced() {
     if [[ ! is_sourced ]]; then
         # We are in a script.
         echo -e "${ERROR}: This needs to be sourced and not executed"
-        exit -1
+        die
```
gpsaggese, line 34: Ok. `die` is a bit ominous, I would have called `abort` or `exit_script`, but it's ok with me.

gpsaggese (commented):
LG. Merging

PR #786, opened 2025-06-01
HelpersTask762_Clean_up_the_purification_layer

sandeepthalapanane:
Addressing #762


helpers/hunit_test.py
```diff
@@ -371,40 +371,102 @@ def filter_text(regex: str, txt: str) -> str:
...
+            "Added CSFY_HOST_GIT_ROOT_PATH '%s' for replacement", csfy_git_root
+        )
+    # Priority 3: Current working directory.
     pwd = os.getcwd()
-    pattern = re.compile(f"{pwd}{dir_pattern}")
-    txt = pattern.sub("$PWD", txt)
-    # 4) Replace the current user name with `$USER_NAME`.
+    if pwd and pwd != "/" and pwd not in processed_paths:
+        replacements.append((pwd, "$PWD"))
+        processed_paths.add(pwd)
```
gpsaggese, line 435: Is `processed_paths` really needed?
Why not just add the replacements and then remove the ones that are common in a stage?

The code will become simpler

helpers/test/test_hunit_test.py
```diff
@@ -1351,3 +1522,103 @@ def test1(self) -> None:
...
         actual = hunitest.purify_docker_image_name(txt)
         self.assert_equal(actual, expected, fuzzy_match=True)
+
+
+# #############################################################################
+# TestPurifyAppReferences1
+# #############################################################################
+
+
+class TestPurifyAppReferences1(hunitest.TestCase):
```
gpsaggese, line 1532: Let's call this `Test_purify_app_references1` so it's easier to grep for

gpsaggese (commented):
Overall well done. It wasn't an easy PR, and I am afraid it won't be easy to pass across all repos.

I've suggested some improvements and then we can merge and do another round of clean up.

helpers/hunit_test.py
sandeepthalapanane: I don't think we need `processed_paths` or to remove the common ones. If we remove the sorting, we won't break the priority. Once a path is replaced, it won't match any subsequent replacements.

gpsaggese (commented):
Good work. I've made some small cosmetic changes and we can merge

Issue #789, opened 2025-06-02
Move purification code in class / file


gpsaggese:
The purification code has become so complex that we should move it in its own file hunit_test_purification.py and convert it into a class so there is a bit of code cohesion (even if there is no shared state, but at least we can indicate there is some sort of relationship between the functions).

I'll file a follow up issue for this

_Originally posted by @gpsaggese in https://github.com/causify-ai/helpers/pull/786#discussion_r2121025588_

gpsaggese:
All done

PR #841, opened 2025-06-15
HelpersTask789_Move_purification_code_in_class_file

sandeepthalapanane:
Addressing #789

- Moved purification code and its tests to separate scripts.
- Added missing tests for some purify functions.
- Updated all purify calls across the codebase using grep.


gpsaggese (commented):
Good work!

Issue #748, opened 2025-05-21
Clean up the hopenai.py


gpsaggese:
Run the linter 
> i lint --files="helpers/[hopenai.py](http://hopenai.py/),helpers/test/test_hopenai.py"

Follow up on the TODOs and the PR reviewer requests

gpsaggese:
- Add support for openrouter
- Keep OpenAI as default
- Add corresponding unit tests

PR #870, opened 2025-06-22
HelpersTask748_Clean_up_the_hopenai.py_01(Made cost functions in to class)

srinivassaitangudu:
Pre-commit checks:
All checks passed ✅

Issue: #748


helpers/hllm.py
```diff
@@ -273,92 +273,121 @@ def _call_api_sync(
...
-_CURRENT_OPENAI_COST = None
+# #############################################################################
+# LLMCostTracker
+# #############################################################################
 
 
-def start_logging_costs() -> None:
-    global _CURRENT_OPENAI_COST
-    _CURRENT_OPENAI_COST = 0.0
+class LLMCostTracker:
```
gpsaggese, line 284: The goal is to make this class the "state", so this guy should have the state in the init.

Then users create one for all the time that they need to track by passing this to get_completion

Does the design make sense?

helpers/hllm.py
```diff
@@ -273,92 +273,121 @@ def _call_api_sync(
...
-    models_info_file: str,
-    provider_name: str = _PROVIDER_NAME,
-) -> float:
-    """
-    Calculate the cost of an OpenAI API call.
+        :param completion: The completion response from OpenAI
+        :param model: The model used for the completion
+        :return: The calculated cost in dollars
+        """
+        """
```
gpsaggese, line 336: Two times same comment

gpsaggese (commented):
Good work

Issue #746, opened 2025-05-21
Extend simple_cache


gpsaggese:
How to extend simple_cache so that we can cover multiple "uses".

From Sai

```
Regarding not reusing the hcache_simple.py code, I actually spent most of my time reading and understanding it before writing the new code.
There are a couple of reasons why I didn’t use hcache_simple directly:
The cache format is different from what we needed for get_completion().
We primarily handle different cache_modes, and hcache_simple doesn’t support that behavior.
I could have reused functions like force_cache_from_disk() and flush_cache_to_disk(), but I felt it might make the code more confusing.
That said, I can still try to reuse some of the functions where appropriate.
I’d appreciate your insights on this and will adjust my approach accordingly.
```

- [x] Make proposal
- [ ] Consider also how to test

srinivassaitangudu:
@gpsaggese please review the PR #793.

gpsaggese:
Added perms, can you try now https://github.com/orgs/causify-ai/projects/80 ?

srinivassaitangudu:
@gpsaggese I have only view access.

gpsaggese:
Can you try now?

gpsaggese:
What's next on this one? The goal is to get simple_cache to support everything we need for now.

E.g., I have another use case to support. 

ImageHashCache in render_image.py takes an input and generates some text.
The idea is that we want to replace this with simple_cache.
It's probably a trivial replacement.

Can you pls take a look and propose a plan?

gpsaggese:
@srinivassaitangudu to make a proposal

gpsaggese:
This is how I think about this problem.

We want to replace `_CompletionCache` and `ImageHashCache` in `render_image.py` with an improved version of `SimpleCache`.

What are the things that those two classes have that `SimpleCache` doesn't have? IMO
1) Both specialized caches save some extra metadata
    - How does your proposal addresses adding code to the cache output? We need to have a way to inject specialized behaviors in the class
2) Is there anything that `_CompletionCache` has in terms of the functionalities "DISABLE_CACHE","REFRESH_CACHE", "HIT_CACHE_OR_ABORT", "NORMAL"
    - Can cache_simple support these modes, so we can unify?
3) It might be useful to support when the output is a binary
     - This can be done in several ways
        - encoding the binary input in the dict (not great)
        - add a file name as value and instead of returning that one returns the content of the binary file
4) The rest of the use cases IMO are not needed
```
[...]

PR #878, opened 2025-06-24
HelpersTask746_Extend_simple_cache

srinivassaitangudu:
issue : #746

### DIFF
--- a/dev_scripts_helpers/documentation/render_images.py
+++ b/dev_scripts_helpers/documentation/render_images.py
@@ -112,8 +112,7 @@ def _render_image_code(
         inserted
     :param dst_ext: extension of the rendered image, e.g., "svg", "png"
     :param dry_run: if True, the rendering command is not executed
-    :return: path to the rendered image and a boolean indicating if the
-        cache was hit
+    :return: path to the rendered image
     """
     _LOG.debug(hprint.func_signature_to_str("image_code_txt"))
     if image_code_type == "plantuml":


helpers/hcache_simple.py
```diff
@@ -461,6 +461,7 @@ def reset_mem_cache(func_name: str = "") -> None:
 
 
 def reset_disk_cache(func_name: str = "") -> None:
+    # Question(Sai) : Why are we doing "assert 0" ?
```
gpsaggese, line 464: We don't want to blast away the cache by mistake.

You can replace the assert 0 with hsystem.query_yes_no(msg)

dev_scripts_helpers/documentation/render_images.py
```diff
@@ -280,23 +282,23 @@ def _render_image_code(
...
-    if use_cache:
-        # Initialize cache handler.
-        cache_file = cache_file or "./tmp.render_images.cache.json"
-        _LOG.debug(hprint.to_str("cache_file"))
-        cache = ImageHashCache(cache_file)
-        # Compute hash of inputs.
-        cache_key, cache_value = cache.compute_hash(
-            image_code_txt, image_code_type, out_img_file_path
-        )
+    # cache_hit = False
```
gpsaggese, line 285: Let's remove all the crap, including ImageHashCache if not needed.

dev_scripts_helpers/documentation/render_images.py
```diff
@@ -348,9 +219,9 @@ def _render_image_code(
     # Remove the temp file.
     os.remove(in_code_file_path)
     # Update the cache.
-    if use_cache:
-        cache.update_cache(cache_key, cache_value)
-    return out_img_file_path, cache_hit
+    # if use_cache:
+    #     cache.update_cache(cache_key, cache_value)
```
gpsaggese, line 223: What about this?

If it's a leftover, one approach we use is to skim through the PR after sending it out to catch these small imperfections

dev_scripts_helpers/documentation/render_images.py
```diff
@@ -217,8 +99,7 @@ def _render_image_code(
     force_rebuild: bool = False,
     use_sudo: bool = False,
     dry_run: bool = False,
-    cache_file: Optional[str] = None,
-) -> Tuple[str, bool]:
+) -> str:
```
gpsaggese, line 102: The :return part is not updated

dev_scripts_helpers/documentation/render_images.py
```diff
@@ -84,128 +84,10 @@ def _get_rendered_file_paths(
...
-        """
-        Save the hash cache to a file.
-        """
-        hio.to_json(self.cache_file, self.cache)
-
-
 # #############################################################################
 
 
+@hcacsimp.simple_cache(write_through=True)
```
gpsaggese, line 90: Can we explain in a comment why we set write_through here?

If we are overwriting a default param, we should explain it for the principle of minimum surprise

dev_scripts_helpers/documentation/render_images.py
```diff
@@ -84,128 +84,10 @@ def _get_rendered_file_paths(
...
-        return hio.from_json(self.cache_file)
-
-    def _save(self) -> None:
-        """
-        Save the hash cache to a file.
-        """
-        hio.to_json(self.cache_file, self.cache)
-
-
 # #############################################################################
```
gpsaggese, line 89: is use_cache used?

I don't think so, but in this case the linter should have caught this

gpsaggese (commented):
Some small issues. Let's always pay attention to details. The goal is only perfect PRs

srinivassaitangudu:
> Some small issues. Let's always pay attention to details. The goal is only perfect PRs

Thanks for the feedback! Just to clarify — I had requested a review for PR #886 , so I think there might’ve been a mix-up. That said, I’m still cleaning up the code and updating the test cases on this one, and I’ll make sure everything is polished. Appreciate the reminder to focus on the details — will definitely keep that in mind.

PR #885, opened 2025-06-25
Update install_os_packages.sh

samarth9008:
To install `pygraphviz` package

FYI @heanhsok


heanhsok:
- Maybe we can also make this specific to the downstream runnable directory (e.g., remove the link, replace with the actual file, and customize it if it's needed) as it might not be needed globally (i.e. in all the runnable dirs). The reason is that we want to keep the number of base deps small
- As discussed in https://github.com/causify-ai/cmamp/pull/12576#discussion_r2166942508, the `pygraphviz` might also not be needed
- I can file follow ups to work on this 

LMK what you think @samarth9008

PR #886, opened 2025-06-26
Helpers task746 extend simple cache 02

srinivassaitangudu:
issue : #746


helpers/hcache_simple.py
```diff
@@ -491,19 +491,43 @@ def decorator(func: Callable[..., Any]) -> Callable[..., Any]:
...
+            *args: Any,
+            **kwargs: Any,) -> Any:
             # Get the function name.
             func_name = func.__name__
             if func_name.endswith("_intrinsic"):
                 func_name = func_name[: -len("_intrinsic")]
             # Get the cache.
             cache = get_cache(func_name)
+
+            if func_name == "_call_api_sync_cached":
```
gpsaggese, line 509: This seems custom? I'm ok if there is a general case and we can decorate the function with a special thing.

A better alternative is to use an option through the interface, instead of having the code infers stuff from the name.

`_intrinsic` is kind of ok since it's a general approach, but this one calls for a switch.

helpers/hcache_simple.py
```diff
@@ -491,19 +491,43 @@ def decorator(func: Callable[..., Any]) -> Callable[..., Any]:
...
+
+                if "client" in kwargs:
```
gpsaggese, line 511: Let's add some comments explaining the highl level logic of we are doing and why

helpers/hcache_simple.py
```diff
@@ -491,19 +491,43 @@ def decorator(func: Callable[..., Any]) -> Callable[..., Any]:
...
+                        # Abort if the cache is not hit.
+                        _LOG.debug("Abort on cache miss")
+                        abort_on_cache_miss = True
             # Get the key.
-            # key = (args, frozenset(kwargs.items()))
-            key = args
-            key = str(key)
+            key = str({"args":args, "kwargs": kwargs})
             _LOG.debug("key=%s", key)
+            # Handling hllm.get_completion() calls.
```
gpsaggese, line 529: ? This is specific of the caller. There should be no reference from the decorator to the decorated.

helpers/hllm.py
```diff
@@ -265,8 +265,33 @@ def _call_api_sync(
...
+    cost = _calculate_cost(completion, model, models_info_file)
+    _accumulate_cost_if_needed(cost)
+    # Convert OpenAI completion object to DICT.
+    completion_obj = completion.to_dict()
+    # Store the cost in the completion object.
+    completion_obj["cost"] = cost
+    return completion_obj
+
+
+@hcacsimp.simple_cache(cache_type="json", write_through=True)
```
gpsaggese, line 279: Let's explain why we set the switches as we do.

helpers/hllm.py
```diff
@@ -265,8 +265,33 @@ def _call_api_sync(
         temperature=temperature,
         **create_kwargs,
     )
-    model_response = completion.choices[0].message.content
-    return model_response, completion
+    # Calculate the cost.
+    models_info_file = ""
```
gpsaggese, line 269: I've proposed in a different PR to pass the LlmCost object to get_completion but this will make caching difficult.
You can file a bug for this issue and we can think about how to do it

gpsaggese (commented):
Let's make sure any PR even if it's a draft follows our conventions (e.g., explaining how and why). We need to get to writing beautiful code, even if it's pseudo code. We can cut corners on testing, implementation, etc., but we need to explain the code for ourselves and for the reviewer

Issue #882, opened 2025-06-25
Add tests for purify text to test app.amp references


sandeepthalapanane:
- For 4) I think we can add some unit tests in //helpers using mocked inputs from different scenarios and run them through the `purify_txt_from_client` function to verify the intended output

_Originally posted by @heanhsok in https://github.com/causify-ai/helpers/pull/853#discussion_r2155173656_

FYI @gpsaggese @heanhsok

gpsaggese:
Added project. @sandeepthalapanane do you have access to the projects?

sandeepthalapanane:
> Added project. [@sandeepthalapanane](https://github.com/sandeepthalapanane) do you have access to the projects?

I only have access to the Intern project.

gpsaggese:
Ok I'll file an issue to improve the permissions, but GH is not great here

sandeepthalapanane:
Merged to master #887.

PR #887, opened 2025-06-27
HelpersTask882_Add_tests_for_purify_text_to_test_app.amp_references

sandeepthalapanane:
#882

### DIFF
--- a/helpers/test/test_hunit_test_purification.py
+++ b/helpers/test/test_hunit_test_purification.py
@@ -150,7 +150,7 @@ dev_scripts/test/Test_linter_py1.test_linter1/tmp.scratch/input.py:3: error: Nam
     def test11(self) -> None:
         """
         Test the correct order of `app` -> `amp` purification with multiple
-        references.
+        import statements.
         """
         txt = """
         import app.amp.helpers_root.helpers.test.test_file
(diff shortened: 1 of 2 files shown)


helpers/test/test_hunit_test_purification.py
```diff
@@ -147,6 +147,41 @@ def test10(self) -> None:
...
+        import app.amp.helpers.config
+        from amp.app.helpers.config import get_config
+        import amp.app.helpers_root.config
+        """
+        exp = """
+        import helpers.test.test_file
+        from helpers.hprint import dedent
+        import helpers.config
+        from helpers.config import get_config
+        import amp.app.helpers_root.config
```
heanhsok, line 167: Shouldn't this be purified to `helpers_root.config`? Could you help recheck the logic of the implementation?

helpers/test/test_hunit_test_purification.py
```diff
@@ -147,6 +147,41 @@ def test10(self) -> None:
...
+        """
+        self.check_helper(txt, exp)
+
+    def test12(self) -> None:
+        """
+        Test amp and app references in the same string.
```
heanhsok, line 173: Could we make the docstring clearer for both test cases? For example, the `test11` verifies the `import` statement string and `test12` verifies the file path string.

heanhsok (changes requested):
Some clarifications and nits.

helpers/test/test_hunit_test_purification.py
sandeepthalapanane:
```python
(r"app\.amp\.helpers_root\.helpers", "amp.helpers"),
````

This pattern in the `purify_app_references()` matches `app.amp.helpers_root` but not `amp.app.helpers_root`.

heanhsok (approved):

Issue #892, opened 2025-06-27
Extend render_images to use absolute references


aangelo9:
On https://github.com/causify-ai/helpers/issues/847#issuecomment-3010742890.

Currently, `render_images.py` generates images and inserts them using relative paths. This works well for local or internal rendering workflows but breaks when the output is hosted outside the repo context, such as on blogs or external Markdown renderers (e.g., Ghost).

To improve compatibility with hosted content platforms, we propose adding a `--use_github_hosting` flag that switches rendered image references to use absolute URLs pointing to the raw image content hosted on GitHub (e.g., `https://raw.githubusercontent.com/...`).

FYI @gpsaggese

aangelo9:
Done.

Issue #847, opened 2025-06-16
Document process for publishing on blog


gpsaggese:
We want to document:
1) What is the authorship policy for blog entry
    - It's everyone who helped (e.g., writing code, reviewing code, writing the blog, etc)
 
2) In practice we want to have a simple system where one of our markdown docs becomes a blog entry
    - Pros
       - Helps with making sure people spend effort documenting nicely (e.g., with plots)
       - Use Diataxis
       - We can source control the blog entry (e.g., with a ping back to the repo and hash) so if the documentation is updated also the blog is

Next steps
- [ ] The blog can be called "Causify Research Blog"
- [ ] Use tags to categorize blog entries
- [ ] Come up and document with a process to "push a md to the blog"
- [ ] We want to push the 4-5 that we have ready
- [ ] Once we have enough meat, we'll start cross-posting (on LinkedIn, Substack)

Shayawnn:
- The [Causify Blog Platform Documentation](https://docs.google.com/document/d/1b6MwjEzS-KunCRvHvhebOpySS50IduNrAQWI4zltu9s/edit?usp=sharing) outlines how to log in, create content, invite staff, and maintain the Causify Blog.

gpsaggese:
Perms given

heanhsok:
@aangelo9 We sent you a contributor invite via email. Could you try to sign up? You'll be able to create and edit your own posts, but to publish, you'll need approval.

aangelo9:
## Problem
I have encountered a problem where linked text from the Markdown appears in black, which blends in with the black background of the website. From my understanding, only admin/staff can change the blog format.

![Image](https://github.com/user-attachments/assets/1b8045d1-c120-4b31-8db5-5388905bbca8)

![Image](https://github.com/user-attachments/assets/c4fe71b5-adc6-4723-bc3c-fc10e52f2c05)

## Solution
- Remove TOC and any linked text in the file temporarily
- Edit the theme & design of the blog

WDYT? @heanhsok

aangelo9:
## Problem
For the authorship policy for blog entry, Ghost can only have one author per post.

## Solution
We could add an Authorship section at the end of the blog.

```markdown
## Authorship

- gpsaggese – developed the supporting code and workflows  
- aangelo9 – authored the documentation and blog post
- Shayawnn, heanhsok – managed the publishing process, reviewed the documentation and blog content  
```

[...]

PR #893, opened 2025-06-27
HelpersTask892_Extend_render_images_to_use_absolute_references

aangelo9:
Task #892.

The image `docs/tools/figs/blog_diagrams/invoke_git_branch_copy_diagram.png` is to address https://github.com/causify-ai/helpers/issues/847#issuecomment-3010742890, where we want to push the rendered image to our github repo so that the blog can display the graph using absolute reference.


dev_scripts_helpers/documentation/test/test_render_images.py
```diff
@@ -549,10 +553,6 @@ def test_mermaid3(self) -> None:
         Check bare mermaid code in a LaTeX file.
         """
         in_lines = r"""
```
gpsaggese, line 555: I think the linter is messing this up. Can you revert the changes and file a bug with a repro case?

gpsaggese (commented):
Code is good. Let's revert a change from the linter and merge!

dev_scripts_helpers/documentation/test/test_render_images.py
aangelo9: Bug issue on #900.

PR #894, opened 2025-06-29
Helpers task802 create diataxis documentation templates for llms

aver81:
@gpsaggese 

Based on our previous discussion, I modified the template to include some HTML comment tags to make it easy for LLMs to generate template, compared to the previous version of the template.

I remember we discussed using XML tags but that might make it more difficult for LLMs to properly adhere to the template while adding content. Plus, XML tags should ideally be used when we have a custom XML based pipeline or structure for documentation. HTML tags are more Markdown-native for LLMs, which is why I think this might be a better option. 

Let me know what you think. If good, I'll push the other templates too.


docs/tools/all.template_explanation.md
```diff
@@ -10,46 +11,56 @@
...
 
 ## Core Concepts
 
-- List and explain key ideas, building blocks, or theoretical foundations.
-  - **Concept 1:** Brief explanation
-  - **Concept 2:** How it relates to others
-  - **Concept 3:** Dependencies or assumptions
-- Attach diagrams or charts (e.g., using Graphviz dot, Mermaid, Tikz) whenever
-  possible or necessary.
+List and explain key ideas, building blocks, or theoretical foundations.
```
gpsaggese, line 25: Let's always use bullet lists even for level 0, so you can revert this and the following

docs/tools/template._for_explanation.md
```diff
@@ -0,0 +1,88 @@
...
+Describe the internal logic, architecture, or process.
+Break down each key step or component in a clear, sequential manner.
+Use diagrams or pseudocode where appropriate. Do not include this comment in the final output.
+-->
+1. Step 1: ...
+2. Step 2: ...
+3. Step 3: ...
+
+
+---
```
gpsaggese, line 57: Let's not use the separators

docs/tools/template._for_explanation.md
```diff
@@ -0,0 +1,88 @@
...
+-->
+- Concept 1: ...
+- Concept 2: ...
+- Concept 3: ...
+
+---
+
+[Component A] → transforms → [Component B] → outputs → [Result]
+
+## How It Works
```
gpsaggese, line 44: File name has typo

docs/tools/template._for_explanation.md
```diff
@@ -0,0 +1,88 @@
...
+
+<!--
```
gpsaggese, line 46: The problem is how to consistently use instructions (for humans and LLMs) and examples in a way that doesn't confuses the LLM.
This requires trial and error. 

Let's use the template in a concrete case and see what happens

gpsaggese (commented):
Let's resolve the conflicts, merge and keep moving

gpsaggese:
Can you pls resolve the conflicts so we can merge?

aver81:
@gpsaggese apologies for the delay. Doing that now

PR #902, opened 2025-07-03
HelpersTask891_Handling_assert_0_in_hcache_simple.reset_disk_cache

srinivassaitangudu:
Pre-commit checks:
All checks passed ✅

### DIFF
--- a/helpers/hcache_simple.py
+++ b/helpers/hcache_simple.py
@@ -11,6 +11,7 @@ import pandas as pd
 
 import helpers.hdbg as hdbg
 import helpers.hprint as hprint
+import helpers.hsystem as hsystem
 
 _LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


helpers/hcache_simple.py
```diff
@@ -461,7 +462,9 @@ def reset_mem_cache(func_name: str = "") -> None:
 
 
 def reset_disk_cache(func_name: str = "") -> None:
-    assert 0
+    assert hsystem.query_yes_no(
```
gpsaggese, line 465: Each line should do one thing, so you should run the system query and then assert.

Also I believe query_yes_no has an option to do this automatically if the answer is no.

helpers/test/test_hcache_simple.py
```diff
@@ -121,6 +121,10 @@ def set_up_test(self) -> None:
          - Reset in-memory caches for all cached functions.
          - Set specific cache properties needed for the tests.
         """
+        import helpers.hsystem as hsystem
+
+        # turn off the “are you sure?” prompt
+        hsystem.query_yes_no = lambda *args, **kwargs: False
```
gpsaggese, line 127: Ugh. Better to pass a param to reset_disk_cache like `interactive: bool = True` and then you pass False

PR #926, opened 2025-07-09
CmampTask12709_Fix_bug_in_tmux.py_for_super_repo

heanhsok:
Task https://github.com/causify-ai/cmamp/issues/12709


dev_scripts_helpers/thin_client/tmux.py
```diff
@@ -5,6 +5,26 @@
 
 import logging
 import os
+import sys
+
+# Handle case where the script is not symlinked.
+# We assume the script is in `{repo_root}/dev_scripts_XYZ/thin_client/tmux.py`
+# so the repo root is 3 levels up.
+git_root = os.path.dirname(
```
gpsaggese, line 13: Wouldn't be simpler / more robust to resolve the path
```

```
and then add the including dir to sys.path?

```
real_path = os.path.realpath(__file__)
dir_name = os.path.dirname(real_path)
assert os.path.exists(os.path.join(dir_name, "thin_client_utils.py"), "Can't find ...")
sys.path.append(dir_name)
```

PR #944, opened 2025-07-14
CmampTask12617_Setup_ECS_cluster_and_ECS_task_definition_for_US_Airflow_prod_environment_2

heanhsok:
Task https://github.com/causify-ai/cmamp/issues/12617

- Add invoke task for provisioning of ECS task definition in prod environment

### DIFF
--- a/helpers/lib_tasks_aws.py
+++ b/helpers/lib_tasks_aws.py
@@ -320,11 +320,29 @@ def aws_create_test_task_definition(
     # Register task definition.
     _register_task_definition(task_definition_name, region=region)
 
+@task
+def aws_create_preprod_task_definition(
+    ctx,
+    region: str = hs3.AWS_EUROPE_REGION_1,
+) -> None:
+    """
+    Create a new ECS task definition for preprod environment.
+
+    :param region: region to create the task definition in
+    """
+    _ = ctx
+    hlitauti.report_task()
+    hdbg.dassert_in(region, hs3.AWS_REGIONS)
+    image_name = hrecouti.get_repo_config().get_docker_base_image_name()
+    task_definition_name = f"{image_name}-preprod"
+    # Register task definition.
+    _register_task_definition(task_definition_name, region=region)
+
 
 @task
 def aws_create_prod_task_definition(
     ctx,
-    region: str = hs3.AWS_EUROPE_REGION_1,
+    region: str = hs3.AWS_US_REGION_1,
 ) -> None:
     """
     Create a new ECS task definition.
--- a/helpers/lib_tasks_docker_release.py
+++ b/helpers/lib_tasks_docker_release.py
@@ -1298,6 +1298,8 @@ def docker_release_test_task_definition(
 ):  # type: ignore
     """
     Release candidate image to test ECS task definition.
+    
+    :param region: region to create the task definition in
     """
     hdbg.dassert_in(region, hs3.AWS_REGIONS)
     # Verify that task definition is provided.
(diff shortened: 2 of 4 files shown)


helpers/lib_tasks_docker_release.py
```diff
@@ -1314,13 +1316,43 @@ def docker_release_test_task_definition(
         region=region,
     )
 
+@task
```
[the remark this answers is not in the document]
heanhsok: Yes I will update the doc once the workflow is finalized

heanhsok:
FYI This update should allow for deployment to the new AWS account right now. However, we might want to do some refactoring afterward (mentioned as TODOs in the comments to do) to better handle this support for multi-account, multi-environment, and multi-region deployment.

helpers/haws.py
```diff
@@ -51,7 +51,25 @@ def get_session(
...
             del credentials["aws_s3_bucket"]
-        session = boto3.session.Session(**credentials)
+        # TODO(heanh): Find a better way to handle different AWS accounts
+        # and environments.
+        if aws_profile == "csfy-prod":
+            source_session = boto3.Session(**credentials)
+            sts_client = source_session.client("sts")
+            # TODO(heanh): Use environment variable to get the role ARN.
+            assumed_role = sts_client.assume_role(
+                RoleArn="arn:aws:iam::726416904550:role/ProdAccountAccessRole",
```
Shayawnn, line 61: Let's avoid hardcoding the account ID and get it through a flag passed by the user.

helpers/haws.py
```diff
@@ -51,7 +51,25 @@ def get_session(
...
+                RoleSessionName="csfy-prod-session",
+            )
+            credentials = {
+                "aws_access_key_id": assumed_role["Credentials"]["AccessKeyId"],
+                "aws_secret_access_key": assumed_role["Credentials"][
+                    "SecretAccessKey"
+                ],
+                "aws_session_token": assumed_role["Credentials"]["SessionToken"],
+                "region_name": "us-east-1",
```
Shayawnn, line 70: The region as well, should be managed by a flag.

Shayawnn (changes requested):
Minor changes; Let's pass the account ID and region through flags!

helpers/haws.py
gpsaggese: I'd use an env var since otherwise we need to wire this to all the executables.
If it's fixed can it be a secret?

In general, we should have a way to define all the env vars through some sort of config that is not checked in, but one can easily use to fix all the switches.

Feel free to file an issue if it makes sense to spend some time thinking.

helpers/haws.py
```diff
@@ -51,7 +51,25 @@ def get_session(
         #  then passing everything.
         if "aws_s3_bucket" in credentials:
             del credentials["aws_s3_bucket"]
-        session = boto3.session.Session(**credentials)
+        # TODO(heanh): Find a better way to handle different AWS accounts
+        # and environments.
+        if aws_profile == "csfy-prod":
```
gpsaggese, line 56: What is the diff from csfy-prod and prod now?

Where should we document this?

helpers/lib_tasks_aws.py
```diff
@@ -137,7 +137,11 @@ def release_dags_to_airflow(
 # ECS Task Definition
 # #############################################################################
 
-_AWS_PROFILE = "ck"
+_AWS_PROFILE = {
```
gpsaggese, line 140: Can we add explanation of the profiles here or point to some docs?

gpsaggese (commented):
Big change, should we have a quick PP to understand?
I'm a bit lost...

helpers/haws.py
heanhsok: I filed a separate issue to discuss this: https://github.com/causify-ai/cmamp/issues/12943

helpers/haws.py
heanhsok: - `csfy-prod` is the AWS profile for the new prod account. Similarly, we have `ck` for the current AWS account. I have renamed it to `csfy` instead, so that it's not confused with the environment
- `prod` refers to production environment

helpers/lib_tasks_aws.py
heanhsok: Yes added comments to explain them.

heanhsok:
As this is a major change,  I have extracted this PR into a smaller issue to discuss how we want to handle authentication for this new prod account. We can look at that one first and merge it before this one.

- https://github.com/causify-ai/cmamp/issues/12943

FYI @gpsaggese @Shayawnn

Shayawnn (approved):
Looking good!

PR #953, opened 2025-07-18
Improve_claude_md_instruction

heanhsok:
- Improve `CLAUDE.md` instruction for Claude code


CLAUDE.md
```diff
@@ -34,23 +34,35 @@ The repository uses `pyinvoke` for task automation with a modular task system:
 
 ## Common Development Commands
 
+### Thin environment
+
+```bash
```
gpsaggese, line 39: I usually run it directly in one of my tmux windows so it's already config.

Is this needed?

CLAUDE.md
```diff
@@ -34,23 +34,35 @@ The repository uses `pyinvoke` for task automation with a modular task system:
...
+# Activate thin environment
+source dev_scripts_helpers/thin_client/setenv.sh
+```
+
 ### Testing
 ```bash
```
gpsaggese, line 45: Good stuff.

Maybe we should be more explicit and say "instead of running pytest you need to run..."?
But the container takes a long time to start (I'm going to speed it up).
Also, a good portion of the code base runs even without invoke. 

Humans solve this problem by doing `docker_bash` and running everything inside.

Not sure how to solve this. Maybe worth filing an issue "How to improve unit tests for Claude"?

gpsaggese (commented):
Love it! Let's keep talking about this

gpsaggese:
I made some changes too over the weekend ;-)

CLAUDE.md
heanhsok: Filed https://github.com/causify-ai/helpers/issues/966#issue-3261896866

PR #993, opened 2025-08-14
CmampTask13034_Setup_s3_bucket_for_storing_shared_config_in_the_prod_environment

heanhsok:


helpers/lib_tasks_aws.py
```diff
@@ -144,7 +144,21 @@ def release_dags_to_airflow(
...
+    """
+    Get the shared configs S3 bucket.
+
+    :param environment: environment to get the shared configs for
+    :return: shared configs S3 bucket
+    """
+    if environment in ["prod"]:
+        return "s3://prod-causify-shared-configs"
+    elif environment in ["preprod", "test"]:
+        return "s3://causify-shared-configs"
```
Shayawnn, line 159: Let's fetch through env somewhere in the private repo, and avoid hardcoding (although we already have some hard-coded exposures, but moving forward, let's avoid.)

Shayawnn (changes requested):

helpers/lib_tasks_aws.py
heanhsok: Moved the configs to https://github.com/causify-ai/cmamp/pull/13081

Shayawnn (approved):
Looking good!

PR #1006, opened 2025-08-29
CmampTask7288_Slim_the_dev_image

dremdem:
[#7288](https://github.com/causify-ai/csfy/issues/7288)

- make optional some installation when building the dev image
  - [`dev.Dockerfile`](https://github.com/causify-ai/helpers/blob/master/devops/docker_build/dev.Dockerfile)
    - [Install publishing tools](https://github.com/causify-ai/helpers/blob/98e33af093409f6d0d51bdd492f90faa085f6e49/devops/docker_build/dev.Dockerfile#L51-L53)


helpers/lib_tasks_docker_release.py
```diff
@@ -341,6 +341,8 @@ def docker_build_local_image(  # type: ignore
     just_do_it=False,
     multi_arch="",
     cleanup_installation=True,
+    install_publishing_tools=True,
+    install_aws_cli=True,
```
heanhsok, line 345: 1) IMO, adding this would bloat the function’s interface. Could we instead rely on the repo config for this?
2) How are we separating dev from prod (e.g. what if we want to install for dev but not for prod)?

helpers/lib_tasks_docker_release.py
```diff
@@ -397,6 +401,8 @@ def docker_build_local_image(  # type: ignore
         ("INSTALL_DIND", True),
         ("POETRY_MODE", poetry_mode),
         ("CLEAN_UP_INSTALLATION", cleanup_installation),
+        ("INSTALL_PUBLISHING_TOOLS", install_publishing_tools),
+        ("INSTALL_AWS_CLI", install_aws_cli),
```
heanhsok, line 405: Should we generalize it and call it sth like `install_dev_tools`? This way, we can add move more stuff into it if needed.

heanhsok (changes requested):

helpers/lib_tasks_docker_release.py
heanhsok: Good point. My concern is that we'll be adding a lot arguments and if/else logic throughout the files, especially for OS-level packages. Since we'll be doing this frequently and plus this is gonna be another major change, it's worth it to do some research and implement a good solution for it.

IMO a declarative approach would be much cleaner (i.e., specify what we need in a list and install specified)

Perhaps we could create our own version of requirements files that list all the OS-level packages we need, installing only what's specified (similar to how we handle Python dependencies with pyproject.toml or requirements.txt files). Could you do a brief search and suggest some solutions?

@gpsaggese do you have any suggestion?

helpers/repo_config_utils.py
```diff
@@ -347,6 +347,22 @@ def use_helpers_as_nested_module(self) -> bool:
...
+        value = self._data["runnable_dir_info"].get(
+            "install_publishing_tools", True
+        )
+        return value
+
+    def get_install_aws_cli(self) -> bool:
+        """
+        Return whether the AWS CLI should be installed, True by default.
+        """
+        value = self._data["runnable_dir_info"].get("install_aws_cli", True)
```
heanhsok, line 363: Maybe good to nest this under sth like dependencies, tools, or packages? You can decide a suitable name for it.

heanhsok (changes requested):

helpers/repo_config_utils.py
dremdem: The same thoughts in the: 
https://github.com/causify-ai/helpers/pull/1006#discussion_r2321732609
If we will choose to keep the packages list in the our `repo_config.yml` file.

helpers/lib_tasks_docker_release.py
heanhsok: The approach SGTM. I would use a different name for the file since requirements.txt is already used by python. .toml or .yaml can be good options too.

I suggest obtaining input from the GP first before proceeding

samarth9008: Okay to do it as a follow-up IMO

It seems like I seem to lack bit of a context. @heanhsok @dremdem Could you explain where are we going with all this? What made you have this comment - https://github.com/causify-ai/helpers/pull/1006/files#r2319354912

samarth9008 (commented):
I am not sure if tweaking AWS CLI installation is worth it bcz it is required for both FE and BE. Are there any use cases where it will not be required?

Also from my experience, publishing tool is only required for linter (helpers) image but we will have check it?

helpers/lib_tasks_docker_release.py
dremdem: > Could you explain where are we going with all this?

The main Idea was:
- Instead of managing the installation of some OS packages directly in the Dockerfiles, we will declare all packages in a single file as a list.  
- This way, we don't need to add a lot of arguments in the Dockerfiles and then process them.  
- Every OS package should have its own installation script, probably following a conventional naming scheme, e.g.:  
  - `install_os_publishing_tools.sh`  
  - `install_os_aws_cli.sh`  
- We can have a centralized script, `install_os_packages.sh`, which will install all OS packages listed in the OS dependencies file.

dremdem:
> I am not sure if tweaking AWS CLI installation is worth it bcz it is required for both FE and BE.

Just for my understanding where we are using the AWS CLI in the FE?

samarth9008:
https://github.com/causify-ai/csfy/blob/d6184b4928d5311abd90cbb6416199ed929c3aac/ck_web_apps/itsavvy/app/app.py#L82

dremdem:
> Separate dependencies in docker scripts #1028

GP filed this one: 
https://github.com/causify-ai/helpers/issues/1028

samarth9008:
@dremdem 

If I am not wrong, one more follow-up would be to edit the `repo_config.yaml` files of atleast the Xerox FE app dir and then we will make similar changes to xerox backend image. WDYT?

dremdem:
> If I am not wrong, one more follow-up would be to edit the `repo_config.yaml` files of atleast the Xerox FE app dir and then we will make similar changes to xerox backend image. WDYT?

I’m lost. 
What exact edit should be made in `repo_config.yaml`?

dremdem:
> Whether we want to install publishing tools or aws cli. If I am not wrong we use `repo_config.yaml` to check what tools to install inside dev docker image
> 
> The task should go on csfy repo.

Got it. That’s what [helpers#1028](https://github.com/causify-ai/helpers/issues/1028) is about.  
And yes, it might be a section of `repo_config.yaml`.

dremdem:
> SHould we file a follow-up to do it and implement it?

I think this one https://github.com/causify-ai/helpers/issues/1028
 is enough, isn’t it?

Issue #1010, opened 2025-09-03
Fix LaTeX Unicode character error in Document Toolchain


protocorn:
The [Toolchain ](https://github.com/causify-ai/helpers/tree/master/dev_scripts_helpers/documentation) cannot handle Greek letters or Unicode characters like "σ" (Sigma)

Fix:
**Add Unicode support packages to LaTeX**

```
\usepackage[utf8]{inputenc}
\usepackage{textgreek}
```

PR #1011, opened 2025-09-03
Fix Unicode charcter error

protocorn:
#1010


dev_scripts_helpers/documentation/latex_abbrevs.sty
```diff
@@ -1,5 +1,14 @@
 % From https://tex.stackexchange.com/questions/75667/change-colour-on-chapter-section-headings-lyx
 \usepackage{xcolor}
+
```
gpsaggese, line 3: There are multiple latex_abbrevs in //tutorials and //umd_classes (we should remove these multiple copies, but Latex is horrible at handling dirs). Can you copy / paste the same changes around?

Issue #1023, opened 2025-09-11
Docker Image does not build on MacOS


indrayudd:
On MacOS while trying to build a docker image for a tutorial, the following error occurs:

```
(client_venv.helpers) (base) Indrayudds-MacBook-Air:tutorial_openai indro$ i docker_build_local_image --version 1.1.0
12:49:15 - INFO  hdbg.py init_logger:1037                               > cmd='/Users/indro/src/venv/client_venv.helpers/bin/invoke docker_build_local_image --version 1.1.0'
# docker_build_local_image: version='1.1.0', cache=True, base_image='', poetry_mode='update', container_dir_name='.', just_do_it=False, multi_arch='', cleanup_installation=True
cp -f devops/docker_build/dockerignore.dev /Users/indro/src/tutorials1/.dockerignore
tar -czh . | DOCKER_BUILDKIT=0 \
    time \
    docker build \
     \
    --build-arg AM_CONTAINER_VERSION=1.1.0 --build-arg INSTALL_DIND=True --build-arg POETRY_MODE=update --build-arg CLEAN_UP_INSTALLATION=True \
    --tag causify/tutorial_openai:local-indro-1.1.0 \
    --file devops/docker_build/dev.Dockerfile \
    -
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            BuildKit is currently disabled; enable it by removing the DOCKER_BUILDKIT=0
            environment-variable.

Sending build context to Docker daemon  870.4kB
Error response from daemon: lsetxattr /openai.example.ipynb: xattr "com.apple.provenance": operation not supported
        0.32 real         0.03 user         0.03 sys
```

Proposed fix would be setting `DOCKER_BUILDKIT` to 1 if the platform is MacOS.

gpsaggese:
1) I don't mind building for Mac so that we don't overfit (too much) to Linux and Mac is a build that developers will want us to support

2) The root cause of that error is that MacOs has stronger permission management that Linux

- Thne solution is to fix the perms issue with something along

```
find . -exec xattr -l {} +
find . -exec xattr -c {} +
```

@indrayudd did you try to google / chatgpt the error?

PR #1024, opened 2025-09-11
HelpersTask1023_Docker_Image_does_not_build_on_MacOS 

indrayudd:
Addresses #1023


gpsaggese:
We don't want to have different build systems for different targets. We'll do a single switch.

It should also be dependent on having a way to build all the images automatically all the times (see @heanhsok) otherwise we can't guarantee sanity

PR #1036, opened 2025-09-24
CmampTask7535_Clean_Docker_images_in_the_main_pytest_py

dremdem:
[#7535](https://github.com/causify-ai/csfy/issues/7535)

- Clean up the Docker images after every tests run the main_pytest.py

### DIFF
--- a/helpers/lib_tasks_docker.py
+++ b/helpers/lib_tasks_docker.py
@@ -61,6 +61,45 @@ def docker_images_ls_repo(ctx, sudo=False):  # type: ignore
     hlitauti.run(ctx, f"{docker_exec} image ls {ecr_base_path}")
 
 
+def docker_images_purge() -> None:
+    """
+    Purge all Docker images to free up disk space.
+    
+    This function runs Docker system prune commands to remove:
+    - All stopped containers.
+    - All networks not used by at least one container.
+    - All dangling images.
+    - All dangling build cache.
+    """
+    _LOG.info("Starting Docker image cleanup...")
+    # Get Docker executable configuration.
+    use_sudo = hdocker.get_use_sudo()
+    docker_exec = hdocker.get_docker_executable(use_sudo)
+    try:
+        # Display disk space before cleanup.
+        _LOG.info("Disk space before cleanup:")
+        hsystem.system("df -h", suppress_output=False)
+        # Run Docker system prune to remove containers, networks, and build cache.
+        cmd = f"{docker_exec} system prune -f"
+        _LOG.info(f"Running: {cmd}")
+        result = hsystem.system(cmd, suppress_output=False)
+        if result != 0:
+            _LOG.warning(f"Docker system prune command failed with exit code {result}")
+        # Run Docker image prune to remove all unused images.
+        cmd = f"{docker_exec} image prune -a -f"
+        _LOG.info(f"Running: {cmd}")
+        result = hsystem.system(cmd, suppress_output=False)
+        if result != 0:
+            _LOG.warning(f"Docker image prune command failed with exit code {result}")
+        # Display disk space after cleanup.
+        _LOG.info("Disk space after cleanup:")
+        hsystem.system("df -h", suppress_output=False)
+        _LOG.info("Docker image cleanup completed")
+    except Exception as e:
+        _LOG.error(f"Error during Docker cleanup: {e}")
+        pass
+
+
 @task
 def docker_ps(ctx, sudo=False):  # type: ignore
     # pylint: disable=line-too-long
--- a/main_pytest.py
+++ b/main_pytest.py
@@ -18,8 +18,29 @@ import helpers.hdbg as hdbg
 import helpers.hgit as hgit
 import helpers.hparser as hparser
 import helpers.hpytest as hpytest
+import helpers.lib_tasks_docker as hlitadoc
 
 _LOG = logging.getLogger(__name__)
+
+
+def _add_common_test_arguments(parser: argparse.ArgumentParser) -> None:
+    """
+    Add common arguments shared by all test commands.
+
+    :param parser: The parser to add arguments to
+    """
+    parser.add_argument(
+        "--dir",
+        action="store",
+        required=False,
+        type=str,
+        help="Name of runnable dir",
+    )
+    parser.add_argument(
+        "--purge-docker-images",
+        action="store_true",
+        help="Purge all Docker images after running tests",
+    )
 
 
 def _parse() -> argparse.ArgumentParser:
@@ -32,35 +53,17 @@ def _parse() -> argparse.ArgumentParser:
     run_fast_tests_parser = subparsers.add_parser(
         "run_fast_tests", help="Run fast tests"
     )
-    run_fast_tests_parser.add_argument(
-        "--dir",
-        action="store",
-        required=False,
-        type=str,
-        help="Name of runnable dir",
-    )
+    _add_common_test_arguments(run_fast_tests_parser)
     # Add command for running slow tests.
     run_slow_tests_parser = subparsers.add_parser(
         "run_slow_tests", help="Run slow tests"
     )
-    run_slow_tests_parser.add_argument(
-        "--dir",
-        action="store",
-        required=False,
-        type=str,
-        help="Name of runnable dir",
-    )
+    _add_common_test_arguments(run_slow_tests_parser)
     # Add command for running superslow tests.
     run_superslow_tests_parser = subparsers.add_parser(
         "run_superslow_tests", help="Run superslow tests"
     )
-    run_superslow_tests_parser.add_argument(
-        "--dir",
-        action="store",
-        required=False,
-        type=str,
-        help="Name of runnable dir",
-    )
+    _add_common_test_arguments(run_superslow_tests_parser)
     parser = hparser.add_verbosity_arg(parser)
     return parser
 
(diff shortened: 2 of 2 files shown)


main_pytest.py
```diff
@@ -110,6 +117,9 @@ def _run_test(runnable_dir: str, command: str) -> bool:
     result = subprocess.run(
         f"invoke {command}", shell=True, env=env, cwd=runnable_dir
     )
+    # Purge Docker images after each test run if requested.
+    if purge_docker_images:
+        hlitadoc.docker_images_purge()
```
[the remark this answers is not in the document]
dremdem: Sounds reasonable.  
Reworked to delete only the images that were used to run the tests.  
Added a flag to delete images that are disabled by default.  
Another option is to delete them anyway when running inside CI.  

WDYT @heanhsok?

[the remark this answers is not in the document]
heanhsok: I think it's better to limit the side effect to just the image in question (e.g. the image of the runnable dir) since pruning everything will not just delete the image that our system download but possibly other images on the system as well

main_pytest.py
```diff
@@ -18,10 +18,61 @@
...
 import helpers.hparser as hparser
 import helpers.hpytest as hpytest
+import helpers.hserver as hserver
+import helpers.lib_tasks_docker as hlitadoc
+import helpers.repo_config_utils as hrecouti
 
 _LOG = logging.getLogger(__name__)
 
 
+def _get_docker_image_for_runnable_dir(
```
heanhsok, line 28: - What if we implement an invoke target (e.g. `invoke docker_clean_image` so we can run this command from the runnable dir level)? 
- This way, we don't have to implement this logic to find out all these repo configs.
- And we can run this command on our own when needed and limit the side effect to just runnable dir

main_pytest.py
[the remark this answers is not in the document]
dremdem: >  (e.g. the image of the runnable dir)

Yes, I did it like that.  
I also made it so that images are removed by default when running inside CI.  
These changes were made to avoid updating all the GH workflow files.

dremdem:
@heanhsok 
Can you pls review it one more time since it's blocker for the:
https://github.com/causify-ai/csfy/issues/7475
I checked this changes in the:
https://github.com/causify-ai/csfy/pull/7515

helpers/lib_tasks_docker.py
```diff
@@ -61,6 +61,50 @@ def docker_images_ls_repo(ctx, sudo=False):  # type: ignore
...
+        # Check if the image exists first and get size info.
+        check_cmd = f"{docker_exec} images --format 'table {{{{.Repository}}}}:{{{{.Tag}}}}\\t{{{{.Size}}}}' {image_name}"
+        _, output = hsystem.system_to_string(check_cmd)
+        if (
+            not output.strip()
+            or "REPOSITORY" in output
+            and len(output.strip().split("\n")) <= 1
+        ):
+            _LOG.info("Image %s not found locally, skipping deletion", image_name)
+            return
```
heanhsok, line 85: - What if we just run rm right away, put it in a try/catch, and show the error message from Docker? For example, 
```
❯ docker rmi -f random
Error response from daemon: No such image: random:latest
```
- This way, we don't have to reimplement any logic to check if the image exists

main_pytest.py
```diff
@@ -110,6 +156,17 @@ def _run_test(runnable_dir: str, command: str) -> bool:
     result = subprocess.run(
         f"invoke {command}", shell=True, env=env, cwd=runnable_dir
     )
```
heanhsok, line 158: And the invoke command would go here. For example,

```bash
if purge_docker_images:
  result = subprocess.run(
        f"invoke docker_clean_image", shell=True, env=env, cwd=runnable_dir
  )
```

main_pytest.py
```diff
@@ -156,6 +216,25 @@ def _main(parser: argparse.ArgumentParser) -> None:
...
+    elif hserver.is_inside_ci():
+        # In CI - purge by default.
+        purge_docker_images = True
+        _LOG.info("Docker image cleanup enabled (running in CI)")
+    else:
+        # Not in CI and flag not specified - don't purge.
+        purge_docker_images = False
+        _LOG.info(
+            "Docker image cleanup disabled (not in CI, use --purge-docker-images to force)"
+        )
```
heanhsok, line 237: And instead of implementing this logic, can we just pass flag when running test from CI? Maybe something link
```bash
./helpers_root/main_pytest.py run_fast_tests --remove-docker-image-after-tests
```

heanhsok (changes requested):

heanhsok:
Also, @dremdem 

To unblock others, we can implement a temporary quick hack in another PR that checks if we are running from CI and performs a system prune all in the following function.

```
def _run_test(runnable_dir: str, command: str) -> bool:
    """
    Run test in for specified runnable directory.

    :param runnable_dir: directory to run tests in
    :param command: command to run tests (e.g. run_fast_tests,
        run_slow_tests, run_superslow_tests)
    :return: True if the tests were run successfully, False otherwise
     """
....
```

For example,
```bash
if running_from_ci:
   hsystem.system("docker system prun -a")
```

main_pytest.py
dremdem: I have added a flag here:  
https://github.com/causify-ai/helpers/blob/d6e5227ba7986d82d98c8be045bea702197b3cf1/main_pytest.py#L69-L73  

As I mentioned [in the comment](https://github.com/causify-ai/helpers/pull/1036#discussion_r2379500673), I made the images be removed by default when running inside CI in order to not add this flag to every GH workflow file. 

WDYT?

main_pytest.py
dremdem: Done.

helpers/lib_tasks_docker.py
dremdem: Done.

main_pytest.py
```diff
@@ -112,10 +117,19 @@ def _run_test(runnable_dir: str, command: str) -> bool:
...
-    # Temporarily hack to clean up docker images after running tests.
-    # See CmampTask7535.
-    if hserver.is_inside_ci():
-        hsystem.system("docker system prune -a -f")
+    # Clean up the Docker image used in the test run if requested.
+    if purge_docker_images:
+        _LOG.info("Cleaning up Docker image")
+        # Display disk space before cleanup.
+        _LOG.info("Disk space before cleanup:")
+        subprocess.run("df -h", shell=True)
```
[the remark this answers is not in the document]
dremdem: Done.

main_pytest.py
```diff
@@ -162,6 +173,25 @@ def _main(parser: argparse.ArgumentParser) -> None:
     hdbg.init_logger(verbosity=args.log_level, use_exec_path=True)
     command = args.command
     runnable_dir = args.dir
+    # Determine Docker image cleanup behavior:
+    # - If in CI: cleanup by default.
+    # - If not in CI: don't cleanup unless flag is specified.
+    # - If flag is specified: force cleanup regardless of CI status.
```
heanhsok, line 179: Can we align the order of the comments with the order of the if/else statement? Probably not needed too, since you already have the comments under the ifelse line.

main_pytest.py
```diff
@@ -112,10 +117,13 @@ def _run_test(runnable_dir: str, command: str) -> bool:
...
-    # Temporarily hack to clean up docker images after running tests.
-    # See CmampTask7535.
-    if hserver.is_inside_ci():
-        hsystem.system("docker system prune -a -f")
+    # Clean up the Docker image used in the test run if requested.
+    if purge_docker_images:
+        _LOG.info("Cleaning up Docker image")
+        # Delete the Docker image (disk space reporting is now handled by the task itself).
+        result = subprocess.run(
+            f"invoke docker_remove_image", shell=True, env=env, cwd=runnable_dir
```
heanhsok, line 125: Can we make the naming consistent? ATM I see we're using `remove` and `purge` in different places. I suggest using `remove` everywhere

helpers/lib_tasks_docker.py
```diff
@@ -61,6 +61,45 @@ def docker_images_ls_repo(ctx, sudo=False):  # type: ignore
...
 
+@task
+def docker_remove_image(ctx, base_image="") -> None:  # type: ignore
+    """
+    Delete the current dev image to free up disk space.
+
+    :param base_image: base name of the image (e.g., `*****.dkr.ecr.us-
+        east-1.amazonaws.com/amp`)
+    """
+    # - Display disk space before cleanup.
```
heanhsok, line 72: We don't need bullet points (`-`) in front of comment.

heanhsok (changes requested):

helpers/lib_tasks_docker.py
```diff
@@ -61,6 +61,39 @@ def docker_images_ls_repo(ctx, sudo=False):  # type: ignore
...
+    _LOG.info("Deleting Docker image: %s", image)
+    # - Get Docker executable configuration.
+    use_sudo = hdocker.get_use_sudo()
+    docker_exec = hdocker.get_docker_executable(use_sudo)
+    # - Delete the specific image.
+    cmd = f"{docker_exec} rmi -f {image}"
+    _LOG.info("Running: %s", cmd)
+    try:
+        result = hsystem.system(cmd, suppress_output=False)
+        if result != 0:
```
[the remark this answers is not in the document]
heanhsok: Good point. I'm thinking we can tag the image `dev` as `dev-backup` before deletion and add another invoke to restore the image by retagging `dev-backup` as `dev`. WDTY?

main_pytest.py
dremdem: Resonable. 
Removed.

helpers/lib_tasks_docker.py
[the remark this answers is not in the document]
heanhsok: - I was referring to a scenario where someone might accidentally delete an image or the deletion fails partly (causing the image to be corrupted), and they want to rollback/recover it without rebuilding the whole thing.
- You're right! If we're keeping a backup copy, then we're not freeing up the disk space needed.
- Since the image is stored in a remote registry, we can always recover from there, too.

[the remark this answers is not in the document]
dremdem: > So maybe we don't need to do anything here since rerunning the tests will pull the new image from the remote registry ?

Agreed.

heanhsok (approved):
Feel free to merge

Issue #1037, opened 2025-09-25
Update the docs related to new declarative approach for installing OS packages


dremdem:
Follow-up from the #1028

https://github.com/causify-ai/helpers/pull/1035#issuecomment-3334567224

# Description 

We need to document how to use the declarative approach for installing OS packages when building the dev image.  
The changes are available in https://github.com/causify-ai/helpers/pull/1035

Docs to update: 
- [Docker docs](/docs/tools/dev_system/all.devops_docker.how_to_guide.md)
- [Runnable dirs docs](/docs/tools/dev_system/all.create_a_runnable_dir.how_to_guide.md)

PR #1039, opened 2025-09-29
CmampTask1037_Update_the_docs_related_to_new_declarative_approach_for_installing_OS_packages

dremdem:
#1037

- Update the docs related to new declarative approach for installing OS packages
- remove redundant line in the `os_packages.txt`

### DIFF
--- a/helpers/lib_tasks_docker.py
+++ b/helpers/lib_tasks_docker.py
@@ -62,6 +62,45 @@ def docker_images_ls_repo(ctx, sudo=False):  # type: ignore
 
 
 @task
+def docker_remove_image(ctx, base_image="") -> None:  # type: ignore
+    """
+    Delete the current dev image to free up disk space.
+
+    :param base_image: base name of the image (e.g., `*****.dkr.ecr.us-
+        east-1.amazonaws.com/amp`)
+    """
+    # Display disk space before cleanup.
+    _LOG.info("Disk space before cleanup:")
+    hsystem.system("df -h", suppress_output=False)
+    # Handle the image.
+    stage = "dev"
+    version = ""
+    image = get_image(base_image, stage, version)
+    _LOG.info("Deleting Docker image: %s", image)
+    # Get Docker executable configuration.
+    use_sudo = hdocker.get_use_sudo()
+    docker_exec = hdocker.get_docker_executable(use_sudo)
+    # Delete the specific image.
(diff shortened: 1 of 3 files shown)


devops/docker_build/os_packages/os_packages.txt
```diff
@@ -2,4 +2,3 @@ install_os_publishing_tools.sh
 install_os_aws_cli.sh
 install_os_github_cli.sh
 install_os_graphviz.sh
-install_os_publishing_tools.sh
```
heanhsok, line 5: Why delete instead of commenting out to disable?

docs/tools/dev_system/all.create_a_runnable_dir.how_to_guide.md
```diff
@@ -276,18 +279,21 @@ dependencies
 
 2. Modify dependencies list
 
-- Modify `$CSFY_RUNNABLE_DIR/devops/docker_build/install_os_packages.sh` to add
-  or remove OS packages
+- Modify `$CSFY_RUNNABLE_DIR/devops/docker_build/os_packages/os_packages.txt` to
```
heanhsok, line 282: Can we mention the minimal set to get started with creating the runnable dir so that the dev can know which ones are required and which ones are just extras?

heanhsok (changes requested):
LG overall.

Could we also check if this needs update too?

https://github.com/causify-ai/helpers/blob/master/docs/tools/dev_system/all.create_a_super_repo_with_helpers.how_to_guide.md

devops/docker_build/os_packages/os_packages.txt
dremdem: It’s a duplicate of the first line.

docs/tools/dev_system/all.create_a_runnable_dir.how_to_guide.md
dremdem: It really depends.  
We can build the `dev` image without anything; it will work.  

On the other hand, some of the helper libraries need:  
- `install_os_publishing_tools.sh`  
- `install_os_aws_cli.sh`  
- `install_os_github_cli.sh`  

But in general, it will still work.

dremdem: Added.

Issue #1042, opened 2025-09-29
Clean up mkdocs


Shaunak01:
- https://app.asana.com/1/1204522243707972/project/1211138327783762/task/1211494977698516?focus=true

PR #1044, opened 2025-09-29
HelpersTask1042_Clean_up_mkdocs_2

Shaunak01:
Issue: https://github.com/causify-ai/helpers/issues/1042


.github/workflows/publish_mkdocs.yml
```diff
@@ -20,12 +20,13 @@ jobs:
         with:
           python-version: 3.x
       - run: pip install mkdocs mkdocs-material
-      # To see the modules in `helpers`, PYTHONPATH needs to include `helpers`
+      # To see the modules in `helpers`, PYTHONPATH needs to include `helpers`.
```
gpsaggese, line 23: No need for this

gpsaggese (commented):
I don't like moving docs. Can we create a symlink if it's strictly needed

Issue #1057, opened 2025-10-10
Add Git utils in helpers


Shaunak01:

gpsaggese:
- [ ] Apply changes from https://github.com/causify-ai/helpers/pull/1058#pullrequestreview-3353736116

- [ ] Let's come up with a unit test plan (I know it's not prod code, but let's at least test this lightly)
   - Let's use LLM (I have some Claude instructions in `helpers_root/ai.*`)
   - Check with @tkpratardan if I have some notes from me about this
   - I would start with testing end-to-end rather than unit testing everything
   - We will need some mocking of objects to avoid calling GitHub and also a way to disable the cache (that's why using SimpleCache is best to use, since it has a lot of this machinery to enable testing)

- [ ] Let's add some doc for this "component"
   - Again Claude is your friend here
   - We want to describe the "layers", the high-level architecture, etc
   - Let's create `ai.documentation_instructions.md` to encode how we document code

Let's use this work as a way to refine our AI flow. Feel free to file more issues

PR #1058, opened 2025-10-10
HelpersTask1057_Add_Git_utils_in_helpers

Shaunak01:
Issue: https://github.com/causify-ai/helpers/issues/1057

### DIFF
--- a/helpers/github_utils.py
+++ b/helpers/github_utils.py
@@ -122,14 +122,8 @@ class GitHubCache:
         path = os.path.join(self.cache_dir, f"cache.{func_name}.json")
         return path
 
-
-# #############################################################################
-# GitHubAPI
-# #############################################################################
-
 # Create global cache instance.
 _github_cache = GitHubCache()
-
 
 # #############################################################################
 # GitHubAPI


helpers/github_utils.py
```diff
@@ -0,0 +1,2018 @@
...
+
+_LOG = logging.getLogger(__name__)
+
+
+# #############################################################################
+# GitHubCache
+# #############################################################################
+
+
+class GitHubCache:
```
gpsaggese: Probably SimpleCache would have been sufficient.

Can you take a look and see if we can reuse it? If not, let's add a TODO.

We want to reuse the code as much as possible to avoid reinventing the wheel.
That code is well tested and used everywhere

helpers/github_utils.py
```diff
@@ -0,0 +1,2018 @@
...
+    return stats_list
+
+
+@cached_github_function
+def get_issue_comment_datetimes_by_repo_period_intrinsic(
+    client,
+    org: str,
+    repo: str,
+    username: str,
+    since: datetime.datetime,
```
gpsaggese: are these with timezone?

Personally I prefer pd.timestamp in the interfaces.

gpsaggese (commented):
Good work.
I'll merge and propose some follow up in the GH Issue

PR #1060, opened 2025-10-16
CsfyTask7604_Automate_frequent_rebuild_and_release_cycle_of_docker_images_2

dremdem:
[#7604](https://github.com/causify-ai/csfy/issues/7604)

### DIFF
--- a/helpers/lib_tasks_docker_release.py
+++ b/helpers/lib_tasks_docker_release.py
@@ -5,7 +5,6 @@ import helpers.lib_tasks_docker_release as hltadore
 """
 
 import datetime
-import json
 import logging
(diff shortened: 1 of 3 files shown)


helpers/hversion.py
```diff
@@ -217,6 +217,35 @@ def _check_version(code_version: str, container_version: str) -> bool:
     return is_ok
 
 
+def bump_version(version: str, bump_type: str = "minor") -> str:
```
gpsaggese: need a * or let's make it mandatory. Default params just create uncertainty

dremdem: Added *.

PR #1096, opened 2025-11-10
CmampTask7965_Fix_the_triggering_of_the_release_Docker_images_workflow

dremdem:
[#7965](https://github.com/causify-ai/csfy/issues/7965)

- Fix the triggering of the release Docker images workflow and update the token to `GITHUB_TOKEN` only

### DIFF
--- a/.github/workflows/common_dev_image_build.yml
+++ b/.github/workflows/common_dev_image_build.yml
@@ -120,5 +120,6 @@ jobs:
           CSFY_AWS_DEFAULT_REGION: ${{ env.AWS_DEFAULT_REGION }}
           CSFY_AWS_S3_BUCKET: ${{ vars.CSFY_AWS_S3_BUCKET }}
           GH_ACTION_ACCESS_TOKEN: ${{ steps.app-token.outputs.token }}
+          GH_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         # TODO(Vlad): Reviewer for testing puprose, remove before merge.
         # run: invoke docker_build_test_dev_image --reviewers=dremdem


.github/workflows/dev_image_release.yml
```diff
@@ -1,5 +1,10 @@
 name: Dev image release
 on:
+  # Trigger on a merged PR, with restrictions applied at the job level.
+  pull_request:
+    types: [closed]
```
heanhsok: Isn't this CI pipeline going to appear in all the closed PR?

.github/workflows/common_dev_image_release.yml
```diff
@@ -1,12 +1,12 @@
...
-            required: false
-            type: string
-            default: '.'
+  workflow_call:
+    inputs:
+      container-dir-name:
+        description: 'Container directory name - directory where Dockerfile and changelog.txt are located.'
+        required: false
+        type: string
+        default: '.'
```
heanhsok: Could we add some comments explain where and how this is used?

heanhsok (changes requested):

.github/workflows/common_dev_image_release.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       # This is needed to pull the Docker image.
       - name: Login to GHCR
         run: |
-          echo "${{ secrets.GH_ACTION_ACCESS_TOKEN || secrets.GITHUB_TOKEN }}" \
+          echo "${{ secrets.GITHUB_TOKEN }}" \
             | docker login ghcr.io -u ${{ github.actor }} --password-stdin
```
heanhsok: Also could we see how the PR would like when it's filed by the GH app as well?

.github/workflows/common_dev_image_release.yml
dremdem: Done. And added it in the invoke call itself.

.github/workflows/dev_image_release.yml
dremdem: Yes, but there’s an `if` condition here:  
https://github.com/causify-ai/helpers/blob/806d92095295640bee7acf07db763b8094c1aa25/.github/workflows/dev_image_release.yml#L21-L25)

So, the workflow will be triggered, but the job itself won’t run unless the condition is met.

heanhsok: I think let's not have it triggered and shown for non relevant PRs

dremdem: I can't find a way to prevent the GitHub workflow from triggering when we merge a PR with the `Automated build` label.  
The only solution I’ve found is to trigger on any merged or closed PR and then filter them out at the job level.  

Another possible approach could be:
- Keep this GitHub workflow as a **manual run**
- Add an extra manual step in our pipeline to execute this workflow

.github/workflows/common_dev_image_release.yml
dremdem: > I see they are already set as ORG level secrets/vars so they should already be accessible by //helpers right?

I tried to debug in in the ssh-session inside the GitHub workflow run, but they are empty: 
the ssh-session step to see the variable and the secret: 
https://github.com/causify-ai/helpers/blob/23192e7dc9b24d499289d8660bd225c7bd6f0c46/.github/workflows/common_dev_image_build.yml#L31-L36
- check inside the ssh-session: 
```
runner@runnervmw9dnm:~/work/helpers/helpers$ env | grep app
app-id=
runner@runnervmw9dnm:~/work/helpers/helpers$ env | grep private
private-key=
runner@runnervmw9dnm:~/work/helpers/helpers$
```

PR #1197, opened 2026-04-07
Helpers task1186 improve hcache simple

sonniki:
#1186

- Adding s3 support to hcache_simple
- Adding per-function cache configuration support (e.g., to define custom cache location for a specific function in that function's decorator instead of setting it globally)
- Adding tests for new functionalities
- Updating docs and tutorial


gpsaggese:
Good work, sorry for being slow here. Can you pls resolve so we can merge?

Issue #1207, opened 2026-04-21
Debug dind and sibling container issue on containerized workflow 


heanhsok:
> @heanhsok could you ptal at the error below? I think because of this change, it does not work via AirFlow anymore.
> 
> error:
> ```
> Traceback (most recent call last):
>   File "/venv/bin/invoke", line 8, in <module>
>     sys.exit(program.run())
>              ^^^^^^^^^^^^^
>   File "/venv/lib/python3.12/site-packages/invoke/program.py", line 398, in run
>     self.execute()
>   File "/venv/lib/python3.12/site-packages/invoke/program.py", line 583, in execute
>     executor.execute(*self.tasks)
>   File "/venv/lib/python3.12/site-packages/invoke/executor.py", line 140, in execute
>     result = call.task(*args, **call.kwargs)
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/venv/lib/python3.12/site-packages/invoke/tasks.py", line 138, in __call__
>     result = self.body(*args, **kwargs)
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/app/helpers_root/helpers/lib_tasks_print.py", line 92, in print_env
>     henv.env_to_str(
>   File "/app/helpers_root/helpers/henv.py", line 536, in env_to_str
>     msg += get_system_signature()[0] + "\n"
>            ^^^^^^^^^^^^^^^^^^^^^^
>   File "/app/helpers_root/helpers/henv.py", line 498, in get_system_signature
>     txt_tmp = hserver.get_docker_info()
>               ^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/app/helpers_root/helpers/hserver.py", line 735, in get_docker_info
>     docker_needs_sudo_ = docker_needs_sudo()
>                          ^^^^^^^^^^^^^^^^^^^
>   File "/app/helpers_root/helpers/hserver.py", line 645, in docker_needs_sudo
>     assert False, "Failed to run docker"
>            ^^^^^
> AssertionError: Failed to run docker
> ```
> 
> log file: https://test.airflow.causify.ai/dags/test.europe.horizon.itsavvy.run_pipeline/grid?dag_run_id=manual__2026-04-20T20%3A08%3A22.375108%2B00%3A00&task_id=derived_to_base_part_mapping&tab=logs
> 
> @samarth9008 FYI
> 
[...]

PR #1203, opened 2026-04-13
CsfyTask8565_Fix_the_issues_due_to_CmampTask10710_2

heanhsok:
Task https://github.com/causify-ai/csfy/issues/8565

Stack PR on top of https://github.com/causify-ai/helpers/pull/731 with the fixes

PR #1208, opened 2026-04-21
HelpersTask1207_Debug_dind_and_sibling_container_issue_on_containerized_workflow

heanhsok:
Task #1207 
Pre-commit checks:
All checks passed ✅


heanhsok:
Adding this check back (https://github.com/causify-ai/helpers/pull/1203/changes#r3113612752), as we need to ensure it does not cause issues when running on ECS Fargate through Airflow, since ECS Fargate does not support either DinD or sibling containers.

See: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/fargate-security-considerations.html

> - No privileged containers or access - Features such as privileged containers or access are currently unavailable on Fargate. This will affect uses cases such as running Docker in Docker.
> 
> - Limited access to Linux capabilities - The environment in which containers run on Fargate is locked down. Additional Linux capabilities, such as CAP_SYS_ADMIN and CAP_NET_ADMIN, are restricted to prevent a privilege escalation. Fargate supports adding the [CAP_SYS_PTRACE](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#other_task_definition_params) Linux capability to tasks to allow observability and security tools deployed within the task to monitor the containerized application.
> 
> - No access to the underlying host - Neither customers nor AWS operators can connect to a host running customer workloads. You can use ECS exec to run commands in or get a shell to a container running on Fargate. You can use ECS exec to help collect diagnostic information for debugging. Fargate also prevents containers from accessing the underlying host’s resources, such as the file system, devices, networking, and container runtime.
> 
> - Networking - You can use security groups and network ACLs to control inbound and outbound traffic. Fargate tasks receive an IP address from the configured subnet in your VPC.

FYI @PomazkinG  @gpsaggese @Shayawnn

helpers/hserver.py
```diff
@@ -633,6 +633,12 @@ def docker_needs_sudo() -> bool:
     """
     if not has_docker():
         return False
+    # This check is required to ensure it does not cause issues when running on ECS
```
PomazkinG, line 636: I think here we want to return False if we are inside AirFlow.

This is a proxy for "if we are inside AirFlow"
```
if not has_dind_support() and not use_docker_sibling_containers():
```

We used to use `is_inside_ecs_container()` previously for that. However, I am not sure if that still would work. Or maybe we can create a new function that detects if we are running smth via AirFlow

Just leaving a TODO is fine

PomazkinG (approved):
Overall LG since it solves the problem at hand

Leaving a non-blocking comment

Issue #673, opened 2025-05-07
Linter check for the GitHub Action Workflows


dremdem:
# Description 

It will be worth to add the `Linter` check for the GH action workflow files, since it's not to easy to test them. 

Found the blazing-fast tool: 
https://github.com/rhysd/actionlint

Downloaded binary and tested on the PR: 
https://github.com/causify-ai/helpers/pull/663

<img width="1261" alt="Image" src="https://github.com/user-attachments/assets/8e7e0f36-8f38-47b5-8e82-07f691ab3923" />

dremdem:
FYI
@samarth9008 @sonniki @heanhsok

dremdem:
>  what made us file this issue

There was an indentation-related mistake in https://github.com/causify-ai/helpers/pull/663.
It’s always better to include checks like this in the `Linter`.

> how urgent is this?

Low priority.

gpsaggese:
Perfect for a bounty.

gpsaggese:
We'll probably switch to `pre-commit` and then we can easily integrate this

https://github.com/rhysd/actionlint/blob/main/.pre-commit-hooks.yaml

Me5h4x1:
I’m interested in #673 with AI coding assistance disclosed. Current master has pre-commit configured but no actionlint hook. I also read the closed #1240 and the note that contributions require vetting, so I won’t submit a duplicate implementation without admission.

Proposed scope, if accepted: integrate the official pinned actionlint hook with the current linter path; address the existing workflow diagnostics; verify a valid workflow, malformed YAML and invalid expressions; document the invocation and exclusions. I have only inspected source and discussion so far, and am not claiming completed tests or an assigned bounty.

Is there a current vetting/application process for this bounty, and is it still funded? If so, please confirm the amount and whether the accepted reward can be paid as USDT on TRON (TRC-20). The onboarding guide mentions Tether without specifying the network.

support991:
Hi @gpsaggese, is #673 still funded and accepting applications for contributor vetting?

I ran actionlint 1.7.12 against the 14 workflow files at f1704df8 and reproduced 13 findings. These include the unsupported `draft` pull_request activity in gitleaks.yml, `commit_sha` on codecov-action v5, and the folded `if` expression in dev_image_release.yml. This was a static check; I have not run your CI or submitted a PR.

[...]

PR #1219, opened 2026-04-28
Add actionlint workflow linter

AgentGoose32:
## Summary
- add a non-modifying linter action that runs `actionlint` on `.github/workflows/*.yml` and `.yaml` files
- wire the new action into the standard linter action list
- skip non-workflow files so regular lint runs are unaffected

## Validation
- `python3 -m py_compile linters/amp_actionlint.py linters/base.py`
- manual smoke: `_Actionlint.check_if_possible()` reports unavailable when `actionlint` is not installed locally and non-workflow files return no output

Refs #673
/claim #673

### DIFF
--- a/linters/amp_actionlint.py
+++ b/linters/amp_actionlint.py
@@ -1,0 +1,34 @@
+#!/usr/bin/env python
+"""
+Run actionlint on GitHub Actions workflow files.
+"""
+
+import logging
+import os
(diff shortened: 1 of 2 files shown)


gpsaggese:
Due to the amount of AI PRs, we don't accept PRs from unknown users. Sorry.

PR #1257, opened 2026-05-20
CsfyTask8875_merge_hcache_into_hcache_simple

tkpratardan:
Pre-commit checks:
All checks passed ✅

https://github.com/causify-ai/csfy/issues/8875


docs/tools/helpers/all.hcache_simple.explanation.md
```diff
@@ -36,11 +37,9 @@
   [`/helpers/hcache_simple.py`](/helpers/hcache_simple.py).
 
 - `hcache_simple` is a lightweight, decorator-based module designed for
-  individual function caching, offering basic in‑memory and disk storage (via
-  JSON or pickle) with manual management and performance tracking.
-- In contrast to `hcache` which is a robust, global caching solution that
-  supports tagged caches, automatic invalidation, and shared cache directories
-  across multiple functions and users, using advanced tools
+  individual function caching, offering in-memory and disk storage (via JSON or
+  pickle) with manual management, performance tracking, and source-change
```
samarth9008: ..and lightweight source-change..

docs/tools/helpers/all.hcache_simple.explanation.md
```diff
@@ -266,6 +265,77 @@
...
+    fresh results.
+    ```
+  - The cached value is still returned; the user decides whether to clear
+- After a cache miss (i.e., the function actually runs with the new code), the
+  stored `func_hash` is updated to reflect the new source
+
+- Note: Only one hash is stored per function, not per cache entry. This is a
+  lightweight "best effort" approach: if a fresh execution updates the hash,
+  stale cache entries for other argument combinations will no longer trigger
+  warnings.
```
samarth9008: This is concerning, why are we not able to trigger warning for stale entries if the func hash is different. Are we not comparing func hash during cache hit?

helpers/hcache_simple.py
samarth9008, line 1887: Do you think the right design would be to store the initial func hash here?

samarth9008 (commented):
Some questions

1. Do we know do a separate PR for deleting `hcache` related stuff?
2. Are there any changes on csfy side? Request my review when the PR is ready if there are any

docs/tools/helpers/all.hcache_simple.explanation.md
tkpratardan: You're right. The doc note was incorrect. It was describing a bug in the implementation where the hash was being updated on every miss, which would silence warnings for other stale entries. Fixed the bug (hash no longer updates on miss) and updated the doc note.

docs/tools/helpers/all.hcache_simple.explanation.md
tkpratardan: Fixed.

helpers/hcache_simple.py
tkpratardan: That's seems right to me. IMO, gap here is if cache entries arrive externally (e.g., pulled from S3) and the function never ran locally, `stored_hash` is `None` and hits skip the comparison entirely. Storing at decoration time with the same if not `existing_hash` guard as type closes that gap without overwriting on re-import, so cross-session detection is preserved. Implemented.

tkpratardan:
> Do we know do a separate PR for deleting hcache related stuff?

Yes, I'll push a new PR to handle the deletion of hcache stuff. 


> Are there any changes on csfy side? Request my review when the PR is ready if there are any

Yes, tracked here: https://github.com/causify-ai/csfy/pull/8972. Will proceed with the review once the changes here are finalized.

helpers/hcache_simple.py
samarth9008: I am not sure if I got this s3 path. Is it even possible to generate cache without function running locally? and if it has ran locally it must have a cache property file in the disk. Correct me if I am wrong

docs/tools/helpers/all.hcache_simple.explanation.md
samarth9008: So basically all the cache entries will generate a warning if the func is changed because we only calculate hash once during the decoration. Am I right?

helpers/hcache_simple.py
tkpratardan: Thanks for catching. This was an oversight, I didn't verify the S3 comment before pushing. Updated the comment with the correct motivation and added a TODO. Also made some changes to other comments.

docs/tools/helpers/all.hcache_simple.explanation.md
tkpratardan: Yes.

samarth9008 (approved):

Issue #1374, opened 2026-09-09
Bounty plan: linter stage to find public functions that should be private ($100)


aijazsiddique:
## Task

"$100 — Write a script to find private functions" (Easy, P0) from the public bounty sheet:
https://docs.google.com/spreadsheets/d/1QiTCyydNQwftMWj3nTL5jWBqOq3UCziFChF08aRNBcE/edit

Spec from the bounty Gdoc: add a stage of the linter that detects public functions that should be private. Find all functions defined in a file, look for their calls in the code base, and if all the calls are local to the same file, the function should be private (name starting with `_`). Provide a way to find and assert these functions, plus a switch to fix them.

I read the bounty onboarding checklist and the relevant guides. Note that #1361 contains another plan for this same task; per the two-contributor rule I am posting a separate competing plan rather than duplicating that thread. I could not add a row to the sign-up sheet (it requires a Google login), so please treat this issue as the sign-up.

Disclosed AI assistance: this plan and the implementation are prepared with AI coding assistance under the supervision of the account owner.

## Current state

- `linters/` runs per-file `Action`s registered in `linters/base.py` (`_MODIFYING_ACTIONS`, `_NON_MODIFYING_ACTIONS`); each action returns lints in the `file:line: message` format.
- The check is inherently cross-file: whether a function can be made private depends on usages in *other* files. No existing action collects repo-wide symbol usage, so this needs a small amount of new machinery.

## Proposed implementation

1. **New non-modifying action** `linters/amp_check_private_functions.py`, registered in `_NON_MODIFYING_ACTIONS` before `flake8`:
   - For each linted `.py` file, collect module-level public function definitions with `ast` (skipping `_`-prefixed names, dunders, `main`, `__init__.py`, and test code via `liutils.is_test_code`).
   - Resolve the repo root from the linted file and parse all `.py` files under it once per run (skipping hidden dirs and `tmp.scratch`), collecting identifier usages (`ast.Name`, `ast.Attribute`) into a per-name usage map.
   - Flag a function only when every usage of its name lies inside its defining file. Zero usages anywhere is also flagged. Usage detection is exact and static (AST identifier matches only), which keeps false positives low; dynamic references (strings, `getattr`) remain a known limitation of a static check.
   - Lint format: `<file>:<line>: function '<name>' is only used within this file and should be renamed to '_<name>'`.

[...]

PR #1375, opened 2026-09-09
Helpers1374: linter stage to find public functions that should be private

aijazsiddique:
Implements #1374 ($100 bounty task "Write a script to find private functions" from the public bounty sheet).

Disclosed AI assistance: developed with AI coding assistance under the supervision of the account owner.

## Problem

Public module-level functions (names not starting with `_`) that are never referenced outside their defining file should be private, per the bounty spec. The repo had no automated way to detect them.

## Solution

New non-modifying linter action `check_private_functions` (`linters/amp_check_private_functions.py`), registered in `_NON_MODIFYING_ACTIONS` before `flake8`:

- Collects the public module-level functions of each linted `.py` file via `ast`, skipping `_`-prefixed names, dunders, `main`, `__init__.py`, test code, `@task`-decorated functions (called dynamically by `invoke`), and names exported in `__all__`.
- Builds a repo-wide map of identifier usages (`ast.Name` + `ast.Attribute`), parsed once per run.
- Flags a function only when every usage of its name is inside its defining file; functions with no usage anywhere are flagged too (they should be private or removed). Any same-name identifier in another file counts as a usage, so the check prefers false negatives over false positives.
- Lint output: `<file>:<line>: the function 'foo' is only used within this file and should be renamed to '_foo'`.
- `--fix` switch renames flagged functions and their usages within the defining file using exact AST node positions (def sites, `Name` and `Attribute` nodes), so strings/comments are never touched, and re-verifies the file parses afterwards.
- `--search_dirs` allows scanning additional repos, since `helpers` is consumed by other repos: a function used only by another repo is reported by default and should then stay public.

## Verification

- 11 new unit tests in `linters/test/test_amp_check_private_functions.py` (detection, end-to-end action, fix mode, rename edits) all pass.
  Environment note: I don't have the project Docker image locally, so the tests were run in WSL Ubuntu (Python 3.14, pytest 8.4.2) with the repo `conftest.py` active. The test file and linter only use stdlib + existing repo helpers.
- `black --line-length 82 --check` clean on both new files; `flake8 --max-line-length=82 --extend-ignore=E203,W503` clean.
- Trial runs on real repo files: no false positives on `helpers/hstring.py` (all public functions are used externally — independently cross-checked), and renames via `--fix` were applied and verified on a scratch copy.
- Known limitation (documented in the module docstring): usages are found statically, so dynamic references (strings, `getattr`) can cause false negatives; usages are searched within the repo of the linted file by default.

Happy to adjust scope (e.g., also consider module-level constants or class methods, or a pedantic mode).

### DIFF
--- a/linters/amp_check_private_functions.py
+++ b/linters/amp_check_private_functions.py
@@ -1,0 +1,462 @@
+#!/usr/bin/env python
+"""
+Find public functions that are only used within their defining file.
+
+A public module-level function (i.e., a function whose name does not start
+with "_") that is never referenced outside of the file where it is defined
+should be private, meaning that its name should start with "_".
+
+The check is static and conservative:
+
+- definitions and usages are collected through `ast` parsing;
+- a function is flagged only when every usage of its name within the
+  searched code is inside its defining file; a function with no usage at
+  all is flagged as well;
+- any identifier with the same name in another file counts as a usage, so
+  that dynamic references cause false negatives instead of false positives.
+
+Note: usages are searched in the repository of the linted file (or in the
+dirs passed through `--search_dirs`), so a function that is used only by
+other repos is still reported; in that case the function should remain
+public.
+
+# Usage example
+
+- Find the functions that should be private:
+> amp_check_private_functions.py sample_file1.py sample_file2.py
+
+- Search for usages also in other dirs:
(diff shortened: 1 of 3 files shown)


gpsaggese:

Issue #1359, opened 2026-09-09
Bounty plan: add pytest_run_class invoke task


franklincg:
I’d like to work on the $50 “Add invoke pytest_run_class and test” bounty from the public bounty sheet.

I reviewed the current repository before implementation. `helpers/lib_tasks/lib_tasks_find.py` already provides the test-file/class resolution path used by `find_test_class`, while the pytest task module does not currently expose `pytest_run_class`. I also checked open PRs for `pytest_run_class` and found no implementation PR at the time of this plan.

Plan:
1. Reuse the existing class-resolution logic instead of collecting the whole repository.
2. Add `pytest_run_class` as an Invoke task that resolves an exact class to its pytest node id and runs only that target.
3. Support a preview/dry-run path and fail safely on missing or ambiguous classes so the command can never fall back to the full suite accidentally.
4. Keep class resolution separate from pytest command construction and preserve pytest’s exit status.
5. Add focused unit tests for successful resolution/execution, preview, missing/ambiguous matches, and the no-full-suite safety invariant.
6. Run the repository-required focused tests/lint and document the exact validation performed in the PR.

I’ll follow the bounty onboarding, coding, testing, and PR checklist requirements. I understand that the published process allows at most two contributors on a bounty and that payment is awarded only for accepted work; this issue does not assume an award.

The work will be AI-assisted, with the changes reviewed and validated before submission.

bheemreddy181:
I'm interested in the listed $50 `pytest_run_class` bounty, using disclosed OpenAI Codex assistance. I see plans #1356, #1359 and #1380, plus other availability inquiries. Given the two-contributor limit, is there any eligible slot remaining, or a different small Python tooling bounty available?

I inspected `helpers/lib_tasks/lib_tasks_find.py`: `_find_test_files` and `_find_test_class(..., exact_match=True)` already resolve test classes without importing test modules. The public finder also has clipboard behavior that a runner should avoid. The resolver has documented regex limitations, so I would preserve or explicitly test its supported class syntax rather than promise arbitrary Python discovery.

If a slot is available, the proposed delivery is an exact-class Invoke runner with containing-file mode and preview; explicit errors for missing or ambiguous matches; safe command quoting and pytest failure propagation; and regression tests proving unrelated test modules are not collected. I would submit a separate detailed plan and complete the required onboarding and existing-solution research before implementation.

[...]

PR #1383, opened 2026-09-09
Add pytest_run_class invoke task

franklincg:
## Summary

Implements the $50 `pytest_run_class` bounty plan tracked in #1359.

- Reuses the existing exact class-resolution helpers instead of collecting/running the full suite.
- Adds a root Invoke task that resolves exactly one class to a pytest node id.
- Fails before pytest execution on missing or ambiguous classes, preventing accidental full-suite fallback.
- Supports `--dry-run` to preview the exact pytest command.
- Routes execution through the existing Docker pytest runner and returns its exit status.
- Adds focused unit coverage for resolution, missing/ambiguous targets, dry-run, command construction, and exit-status propagation.

## Existing-solution search

Before implementation I checked the current repository for an existing `pytest_run_class` task and found none. The closest reusable path is `helpers/lib_tasks/lib_tasks_find.py`, whose `_find_test_files()` + `_find_test_class(..., exact_match=True)` already produce pytest-compatible class node ids; this PR intentionally builds on that instead of introducing another collector.

## Validation

The implementation and focused tests are included in the branch. The authorized remote host transport is intermittently timing out, so I could not honestly claim a completed local Docker/lint run before submission; CI is requested for authoritative validation and I will address any resulting failures.

AI-assisted implementation was used, with the changes reviewed before submission.

Refs #1359


gpsaggese:

Issue #1393, opened 2026-09-09
Bounty plan: add opt-in native fast path for repeated llm_transform.py runs ($300)


hamed-cell:
## Bounty and contributor slot

I would like to take the second contributor slot for the **$300 “Speed up `llm_transform.py`”** bounty.

I reviewed the existing plan and implementation in #1376 / #1379. That work primarily improves cold Docker image construction. I am targeting a different bottleneck: **repeated warm invocations**, where the current wrapper still performs Docker preparation/lookup, path conversion, container startup, and a second Python process before the transform itself runs.

I understand that Causify allows at most two contributors per bounty and pays only the stronger accepted implementation. This issue does not assume assignment or payment until Causify confirms it.

## Current-state source review

I inspected:

* `dev_scripts_helpers/llms/llm_transform.py`
* `dev_scripts_helpers/llms/dockerized_llm_transform.py`
* `dev_scripts_helpers/llms/test/test_llm_transform.py`
* `dev_scripts_helpers/llms/README.md`
* the bounty onboarding and PR-review requirements
* #1376 and PR #1379

The outer `llm_transform.py` delegates LLM-backed transforms through `_run_dockerized_llm_transform()`.

Even with an already-built image, that path still prepares Docker state, converts paths/mounts, starts a container, and launches `dockerized_llm_transform.py` as a second Python process.

The inner script contains the actual read → prompt → write transform flow and can be exposed as a reusable function without changing prompt semantics.

## Existing-solution / package research

I evaluated the main alternatives requested by the bounty.

### Long-lived Docker container + `docker exec`

Docker supports executing commands inside an already-running container.

This could retain dependency isolation while avoiding container creation on every invocation, but introduces lifecycle, cleanup, naming/concurrency, stale-code and cross-request-state concerns.

### Docker SDK for Python

The Docker Python SDK supports detached containers and `exec_run()`.

It would however introduce an additional host dependency for behavior already available through Docker CLI and the repository's existing Docker helpers.

### Persistent client/server worker

A persistent isolated worker could potentially provide even lower repeated latency.

The trade-off is significantly more complexity: daemon lifecycle, protocol, failure recovery, concurrency and data-isolation concerns.

### Faster models / streaming

These can reduce live model latency but do not address deterministic wrapper/container startup overhead and can change quality or UX characteristics.

### Patch-only model responses

[...]

PR #1379, opened 2026-09-09
perf: slim and instrument llm transform container

Larslllllll:
## Summary

This PR advances #1376 with a measured, backwards-compatible optimization for the private `llm_transform.py` Docker image.

- Removes unused `pandas` from the image. The transform path and `dockerized_llm_transform.py` do not use it; optional pandas imports remain available in unrelated reporting helpers.
- Consolidates the Python dependency installation into one pip layer.
- Adds phase timing logs for image preparation, command preparation, and container/LLM execution.
- Adds a deterministic regression test for the generated Dockerfile.

## Validation

- `python -m py_compile dev_scripts_helpers/llms/llm_transform.py dev_scripts_helpers/llms/test/test_llm_transform.py`
- `git diff --check`
- Fresh Docker build succeeded with the new image.
- Deterministic `-p test` smoke test succeeded without pandas and produced the expected digest.
- Old and new images produced identical output for the same smoke-test input.
- Fresh no-cache build benchmark on the same host: old Dockerfile **44.55 s**, new Dockerfile **19.89 s**. This is a 55.35% reduction in this controlled build; exact CI times will vary.

`invoke lint` could not start in the available environment because repository task loading requires the missing `boto3` package. No live LLM call is used by the tests.

The default CLI behavior, Docker isolation, input/output handling, and prompt behavior remain unchanged. The timing data is emitted through logging, not stdout, so stdout transforms remain safe.

AI assistance was used. The account owner reviewed the change and validation results.

PR #1395, opened 2026-09-09
perf: add native fast path for repeated llm transforms

hamed-cell:
Summary

Implements the second approach proposed in #1393 for the $300 llm_transform.py performance bounty.

The existing competing implementation in #1379 primarily reduces cold Docker image build time. This PR targets a different bottleneck: repeated warm invocations, where the wrapper still pays Docker/container/process startup overhead on every transform.

Keep the existing Docker path as the default.
Add an explicit --native opt-in mode for already-provisioned host environments.
Factor the inner transform into one shared run_transform() implementation used by both Docker and native execution.
Import the native implementation lazily so the default Docker path keeps its existing host dependency behavior.
Add deterministic file-input and stdin regression tests.
Document the performance/isolation trade-off.
Measured result

Using the deterministic test prompt so no network/model latency is included:

warm Docker median: 1.3323 s
native median: 0.4354 s
median warm-start reduction: 67.3%

Docker remains the default because it provides dependency and environment isolation. Native mode is deliberately opt-in.

Validation
Deterministic native file-input path verified.
Deterministic native stdin path verified.
Gitleaks passes on the submission commit.
Fork fast/slow workflows cannot execute because they require upstream AWS/GHCR credentials unavailable to forks; the failure occurs during credential configuration before tests run.

Related issue: #1393

### DIFF
--- a/dev_scripts_helpers/llms/README.md
+++ b/dev_scripts_helpers/llms/README.md
@@ -27,8 +27,11 @@ This directory has no subdirectories.
 
 - Applies predefined or custom LLM transformations to code files, particularly
   useful for code review and refactoring.
-- Executes transformations in Docker containers to isolate dependencies and
-  environment requirements.
+- Executes transformations in Docker containers by default to isolate
+  dependencies and environment requirements.
+- Supports an explicit `--native` mode that runs the same transform in the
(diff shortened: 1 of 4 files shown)


gpsaggese:

